PR #98, opened 2023-03-19
feat: update cli and support vue.js

ysy945:
- update cli

image


wre232114:
Seems like the vue template is missing

PR #128, opened 2023-03-24
feat(template): Add templates for issue and PR

ErKeLost:
close: https://github.com/farm-fe/farm/issues/111

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,15 +1,15 @@
 name: "\U0001F41E Bug report"
 description: Create a report to help us improve
 body:
-  - type: markdown
-    attributes:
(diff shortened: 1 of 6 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,49 @@
...
+body:
+  - type: markdown
+    attributes:
+      value: |
+        **Before You Start...**
+
+        This form is only for submitting bug reports. If you have a usage question
+        or are unsure if this is really a bug, make sure to:
+
+        - You have a local rust environment
```
wre232114, line 12: Rust enviroment is not required. The only requirement is that you must have GLIBC >= 2.32 when using Farm on linux

wre232114 (changes requested):

PR #147, opened 2023-03-28
feat: update sourcemap of vue

### LINKED ISSUE #138: ModuleSystem does not clear all previous cache when HMR
### Steps to reproduce

1. Remove the dependency `<Description />` from `examples/react/src/main.tsx`:
```tsx
import React from 'react';
import { Description } from './comps/description';
import './main.css';

import { Title } from './comps/title';

export function Main() {
  return (
    <>
      <Title />
     {/* <Description /> */}
    </>
  );
}
```

2. change  `examples/react/src/comps/clock/index.tsx`.
3. then add `Description` back. The changes of  `examples/react/src/comps/clock/index.tsx` does not apply

### What is expected?

all changes should applied

The changes of  `examples/react/src/comps/clock/index.tsx` does not apply.

This may be related to https://github.com/farm-fe/farm/blob/main/packages/runtime-plugin-hmr/src/hot-module-state.ts#L82, @ysy945 

### System Info

_No response_

### Any additional comments?

_No response_

ysy945:
**Description:**
- update sourcemap of vue

**BREAKING CHANGE:**
- none

close #138


js-plugins/vue/src/farm-vue-plugin.ts
```diff
@@ -1,19 +1,16 @@
 import fs from 'fs';
-import path from 'path';
 import { parse } from '@vue/compiler-sfc';
 import { JsPlugin } from '@farmfe/core';
-import { handleHmr } from './farm-vue-hmr.js';
-import { StylesCodeCache, CacheDescriptor, LessStatic } from './farm-vue-types.js';
-import { genMainCode } from './generatorCode.js';
-import { error } from './utils.js';
```
wre232114, line 8: .js suffix is needed, cause the .ts file is built to native esm

wre232114 (changes requested):

Nirvana-Jie (approved):

wre232114 (approved):

PR #149, opened 2023-03-30
feat(create-farm): add create-farm

ErKeLost:
**Description:**

add create-farm package to create farm projects for users

**BREAKING CHANGE:**


pnpm-lock.yaml
```diff
@@ -1,289 +1,241 @@
-lockfileVersion: '6.0'
+lockfileVersion: 5.4
```
wre232114, line 1: use pnpm v7 to install dependencies

wre232114 (changes requested):

README.md
```diff
@@ -57,24 +57,19 @@ Farm is a super-fast web building tool written in Rust. Benchmark against other
...
+
+```bash
+$ yarn create farm
+```
 
-```sh
-cd my-farm-project
-# Here you can use your package manager of choice instead of npm
-npm start
+With PNPM:
```
wre232114, line 72: Is there something lost

wre232114 (changes requested):

PR #150, opened 2023-03-30
feat: plugin resolve exports

### LINKED ISSUE #177: Resolve failed when there is no main field in package.json
### Steps to reproduce

Add a pacakge that does not have main field. Or file using `import xxx from '.'`

### What is expected?

Resolve index.js by default

throw a resolve error

### System Info

_No response_

### Any additional comments?

_No response_

wre232114:
**Description:**

**BREAKING CHANGE:**

close #139 
close #177


crates/plugin_resolve/src/resolver.rs
```diff
@@ -42,31 +42,35 @@ impl Resolver {
     base_dir: PathBuf,
     kind: &ResolveKind,
   ) -> Option<PluginResolveHookResult> {
+    println!("base_dir {:?} source {:?}", base_dir, source);
```
wre232114, line 45: remove all `println`

crates/plugin_resolve/tests/browser.rs
```diff
@@ -37,18 +37,18 @@ fn resolve_browser_replace() {
       let cwd = file.parent().unwrap().to_path_buf();
       let resolver = Resolver::new(ResolveConfig::default());
 
-      let resolved = resolver.resolve("module-a", cwd.clone(), &ResolveKind::Import);
```
wre232114, line 40: Preserve the tests of browser field

crates/plugin_resolve/src/resolver.rs
```diff
@@ -195,32 +200,93 @@ impl Resolver {
...
+              return Some(self.get_resolve_result(&package_json_info, resolved_path, kind));
+            }
+          }
+          println!("不存在package json");
+          // println!("try_directory package_path {:?}", package_path);
+          // let package_json_info = package_json_info.unwrap_err();
+          let parts: Vec<&str> = source.split('/').filter(|s| !s.is_empty()).collect();
+          let mut prev_path = String::new();
+          let mut result = Vec::new();
+          for component in parts {
```
wre232114, line 241: We have already traversed the ancestor directory in parent loop, this may be duplicated

crates/plugin_resolve/src/resolver.rs
```diff
@@ -355,25 +444,126 @@ impl Resolver {
...
+              }
+            }
+            Value::Object(current_field_obj) => {
+              // TODO resolve import require type
+              println!("我拿的是 object 类型的 {:?}", current_field_obj);
+              println!("判断当前是什么类型的引入 {:?}", kind);
+              let dir = package_json_info.dir();
+              for (key, value) in current_field_obj {
+                match kind {
+                  ResolveKind::Import => {
```
wre232114, line 491: beside `import` and `require` we need to support more condition exports by default(refer to vite), and make it configurable

crates/core/src/plugin/mod.rs
```diff
@@ -195,6 +195,10 @@ pub enum ResolveKind {
   Entry,
   /// static import, e.g. `import a from './a'`
   Import,
+  /// static exports default
+  Default,
```
wre232114, line 199: The `ResolveKind` is used to distinguish the dependency type. For resolve we should create a new type instead of add unrelated types here

crates/plugin_resolve/src/resolver.rs
```diff
@@ -260,15 +321,17 @@ impl Resolver {
...
   ) -> PluginResolveHookResult {
     if let Ok(package_json_info) = package_json_info {
       let external = self.is_module_external(&package_json_info, &resolved_path);
       let side_effects = self.is_module_side_effects(&package_json_info, &resolved_path);
-
       let resolved_path = self
         .try_browser_replace(package_json_info, &resolved_path)
         .unwrap_or(resolved_path);
-
+      let resolved_path = self
```
wre232114, line 332: Does exports need to replace the same as browser?

PR #156, opened 2023-04-03
chore(create-farm): change framework sequence optimize code

ErKeLost:
**Description:**

**BREAKING CHANGE:**
change framework sequence optimize code

### DIFF
--- a/.changeset/purple-trees-rule.md
+++ b/.changeset/purple-trees-rule.md
@@ -1,0 +1,5 @@
+---
+'create-farm': major
+---
+
+change framework sequence optimize code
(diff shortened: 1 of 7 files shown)


.changeset/purple-trees-rule.md
```diff
@@ -0,0 +1,5 @@
+---
+'create-farm': major
+---
+
+change framework sequence optimize code
```
wre232114, line 5: I think we do not need `major` here, `minor` is enough

wre232114 (changes requested):

wre232114 (approved):

PR #160, opened 2023-04-03
feat: update vue-options

ysy945:
**Description:**
- update vue plugin of options

**BREAKING CHANGE:**
- none

- none

### DIFF
--- a/js-plugins/vue/src/utils.ts
+++ b/js-plugins/vue/src/utils.ts
@@ -5,6 +5,8 @@ import {
   PreProcessors,
   PreProcessorsType,
   outputData,
+  FarmVuePluginOptions,
+  ResolvedOptions,
 } from './farm-vue-types.js';
 
 export function warn({ id, message }: outputData) {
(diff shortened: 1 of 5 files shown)


js-plugins/vue/src/utils.ts
```diff
@@ -54,6 +56,72 @@ export function callWithErrorHandle<
   }
 }
 
+export function isArray(val: any) {
+  return Array.isArray(val);
+}
+
+export function isRegExp(reg: any) {
```
wjq990112, line 63: Perhaps declaring functions in this way can provide better type guarding abilities.

```diff
- export function isRegExp(reg: any) {
+ export function isRegExp(reg: unknown): reg is RegExp {
```

js-plugins/vue/src/utils.ts
```diff
@@ -54,6 +56,72 @@ export function callWithErrorHandle<
   }
 }
 
+export function isArray(val: any) {
```
wjq990112, line 59:
```
- export function isArray(val: any) {
+ export function isArray(val: unknown): val is unknown[] {
```

js-plugins/vue/src/utils.ts
wjq990112: I'll fix it with a new pull request.

PR #161, opened 2023-04-04
feat: support embed sass

### LINKED ISSUE #121: [feature] support embed sass
Integrated embed sass as plugin_sass to support compile `.scss`.

Investigate https://crates.io/crates/sass-embedded

callqh:
close #121


rust-plugins/sass/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+
+impl FarmPluginSass {
+  fn new(_config: &Config, _options: String) -> Self {
+    Self {}
+  }
+}
+
+impl Plugin for FarmPluginSass {
+  fn name(&self) -> &str {
+    "FarmPluginReact"
```
wre232114, line 25: FarmPluginSass

wre232114 (changes requested):

wre232114:
Thanks! Please add a changeset using `npx changeset` for the sass plugin package

wre232114:
> Thanks! Please add a changeset using `npx changeset` for the sass plugin package

And should create a `.github/workflows/rust-plugin-sass-ci.yaml` and download artifacts in `.github/workflows/release.yaml` like plugin_react

rust-plugins/sass/src/lib.rs
```diff
@@ -0,0 +1,47 @@
+#![deny(clippy::all)]
+
+use farmfe_core::{config::Config, plugin::Plugin};
+use farmfe_macro_plugin::farm_plugin;
+use farmfe_toolkit::regex::Regex;
+use sass_embedded::{Options, Sass};
+
+fn exe_path() -> std::path::PathBuf {
+  std::path::PathBuf::from(concat!(env!("CARGO_MANIFEST_DIR")))
+    .join("ext/sass/sass-embedded")
```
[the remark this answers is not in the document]
callqh: This is really a problem and I need to figure out how to solve it

callqh:
done @wre232114

rust-plugins/sass/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+
+  fn transform(
+    &self,
+    param: &farmfe_core::plugin::PluginTransformHookParam,
+    _context: &std::sync::Arc<farmfe_core::context::CompilationContext>,
+  ) -> farmfe_core::error::Result<Option<farmfe_core::plugin::PluginTransformHookResult>> {
+    if param.module_type == ModuleType::Custom(String::from("sass")) {
+      let mut sass = Sass::new(exe_path()).unwrap();
+      let res = sass
+        .compile(&param.resolved_path, Options::default())
```
wre232114, line 59: Does sass support passing `scss code` as args instead of `resolved_path` when compile?

rust-plugins/sass/tests/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+fn test() {
+  fixture!("tests/fixtures/index.scss", |file, _cwd| {
+    let resolved_path = file.to_string_lossy().to_string();
+
+    let config = Config {
+      input: HashMap::from([("button".to_string(), resolved_path.clone())]),
+      ..Default::default()
+    };
+
+    // 调用插件的transform方法，将Sass文件编译为CSS文件
```
wre232114, line 23: Translate to English

wre232114 (approved):

rust-plugins/sass/scripts/setup.mjs
```diff
@@ -0,0 +1,93 @@
...
+}
+
+(async () => {
+  try {
+    await getDartSassEmbedded('./ext/sass', 'sass-embedded');
+  } catch (error) {
+    console.error(error);
+    process.exitCode = 1;
+  }
+})();
```
shulandmimi, line 93: This file will run on the user's machine

1. Will it block Install because of network problems (such as China)

2. It depends on `node`. Is there any compatibility problem without compile (Can a minimum version be given?)

shulandmimi (changes requested):

rust-plugins/sass/scripts/setup.mjs
wre232114: I think we can provide a internal binary as downgrade when download fail

callqh: There is no good way to deal with the first point, unless we provide a downgrade file as mentioned in 5, but then we need to pull a copy of each OS file, otherwise we can't get compatibility with each system. But then it will increase the size of the plugin package.

I think the second point is to restrict the node version >= 16 in package.json, the script itself does not need to be compatible.

I wonder if there is a better way to solve this problem?

callqh: > This file will run on the user's machine
> 
> 1. Will it block Install because of network problems (such as China)
> 2. It depends on `node`. Is there any compatibility problem without compile (Can a minimum version be given?)

There is no good way to deal with the first point, unless we provide a downgrade file as mentioned @wre232114 , but then we need to pull a copy of each OS file, otherwise we can't get compatibility with each system. But then it will increase the size of the plugin package.

I think the second point is to restrict the node version >= 16 in package.json, the script itself does not need to be compatible.

I wonder if there is a better way to solve this problem?

shulandmimi: 1. I prefer to place executables in the packages of each system under the npm folder, so that executables are only downloaded at release time (ci), and the executables are distributed by the image server, reducing network problems.

Of course, this is just a suggestion, and you don't have to follow it.

shulandmimi (approved):

rust-plugins/sass/scripts/setup.mjs
shulandmimi, line 1: Whether this file can be deleted?

shulandmimi (approved):

rust-plugins/sass/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+  let arch = get_arch();
+  let entry_file = if let "win32" = os {
+    "dart-sass-embedded.bat"
+  } else {
+    "dart-sass-embedded"
+  };
+  let cur_dir = env::current_dir().unwrap().to_string_lossy().to_string();
+  // user manually installs related dependencies
+  let manual_installation_path = PathBuf::from(&cur_dir)
+    .join(format!("node_modules/sass-embedded-{os}-{arch}"))
```
wre232114, line 42: Two quetions:
1. Does user need to install sass-embedded into their project
2. Does it work with pnpm

wre232114 (approved):

PR #186, opened 2023-04-10
Feature:initially changed the css-plugin

Nirvana-Jie:
**Description:**

I initially changed the css-plugin.
The next step is to parse sassOption
**BREAKING CHANGE:**

None.
None.


examples/react/farm.config.ts
```diff
@@ -1,5 +1,6 @@
 // change to @farmfe/core/config when resolve support conditional exports
-import { defineFarmConfig } from '@farmfe/core/config';
+import { defineFarmConfig } from '@farmfe/core/dist/config';
```
ErKeLost, line 2: now can use the export path to import the module， can directly `import { defineFarmConfig } from '@farmfe/core/config';`

PR #207, opened 2023-04-15
feat:add test for crate utils

Zourunfa:
**Description:**

After today's clone, I happened to find that there are no test cases for the reactive method in the utils under crites, so I added it. I want to gradually read the project source code and participate in it

**BREAKING CHANGE:**


wre232114:
Thank you!

Zourunfa (commented):
需要自己检查吗

PR #215, opened 2023-04-16
feat(example): add new example template

ErKeLost:
**Description:**

5a8c49d5e40572bd678876a31276dee


wre232114 (approved):

wre232114:
I think make the icons colorful and add some styles to the `Getting Started` would be better

PR #230, opened 2023-04-18
feat: support css modules

### LINKED ISSUE #113: [Feature] css modules
Support css modules. Seems like swc_css has already support css modules. We should take a investigation on it

shulandmimi:
**Description:**

support css modules

**BREAKING CHANGE:**

close #113


crates/compiler/tests/fixtures/css/modules/local/output.js
```diff
@@ -0,0 +1,75 @@
...
+    }
+});
+var __farm_global_this__ = globalThis || window || global || self;
+var farmModuleSystem = __farm_global_this__.__farm_module_system__;
+farmModuleSystem.bootstrap();
+var entry = farmModuleSystem.require("b5d64806");
+
+
+//a1b6e7b5.css:
+ .farm_base {
```
wre232114, line 64: can we generate a unique id for the css module automatically

wre232114 (approved):

PR #238, opened 2023-04-21
chore: do not resolve browser when target env is node

### LINKED ISSUE #234: Config compiler resolves to "browser" main field
### Steps to reproduce

Have a module with the following `package.json` where `main.js` imports `sub.js`:

```json
{
  "name": "resolution-example",
  "version": "1.0.0",
  "main": "main.js",
  "browser": {
    "./sub.js": "./browser.js"
  }
}
```

### What is expected?

Importing `resolution-example` in `farm.config.ts` will yield contents of `sub.js`.

Importing `resolution-example` in `farm.config.ts` will yield contents of `browser.js`.

### System Info

```Shell
System:
    OS: macOS 13.3.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 91.53 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.15.0 - ~/.asdf/installs/nodejs/18.15.0/bin/node
    Yarn: 1.22.17 - ~/.asdf/shims/yarn
    npm: 9.5.0 - ~/.asdf/plugins/nodejs/shims/npm
    Watchman: 2023.03.27.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 112.0.5615.121
    Firefox: 107.0.1
    Safari: 16.4
  npmPackages:
    @farmfe/core: latest => 0.6.3
```

### Any additional comments?

_No response_

wre232114:
**Description:**

**BREAKING CHANGE:**

close #234

### DIFF
--- a/crates/plugin_resolve/src/resolver.rs
+++ b/crates/plugin_resolve/src/resolver.rs
@@ -376,6 +376,12 @@ impl Resolver {
             from_str(package_json_info.raw()).unwrap();
 
           for main_field in &self.config.main_fields {
+            if main_field == "browser" {
+              if self.output.target_env == TargetEnv::Node {
+                continue;
+              }
+            }
+
             if let Some(field_value) = raw_package_json_info.get(main_field) {
               if let Value::Object(_) = field_value {
                 let resolved_path = Some(self.get_resolve_node_modules_result(
@@ -497,6 +503,7 @@ impl Resolver {
     kind: &ResolveKind,
   ) -> Option<String> {
     // resolve exports field
+    // TODO: add all cases from https://nodejs.org/api/packages.html
     let exports_field = self.get_field_value_from_package_json_info(package_json_info, "exports");
     if let Some(exports_field) = exports_field {
       let dir = package_json_info.dir();
@@ -603,6 +610,10 @@ impl Resolver {
     package_json_info: &PackageJsonInfo,
     resolved_path: &str,
   ) -> Option<String> {
+    if self.output.target_env != TargetEnv::Browser {
+      return None;
+    }
+
     let browser_field = self.get_field_value_from_package_json_info(package_json_info, "browser");
     if let Some(browser_field) = browser_field {
       if let Value::Object(obj) = browser_field {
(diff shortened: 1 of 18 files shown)


crates/plugin_resolve/src/resolver.rs
```diff
@@ -376,6 +376,12 @@ impl Resolver {
             from_str(package_json_info.raw()).unwrap();
 
           for main_field in &self.config.main_fields {
```
oblador, line 378: FWIW, I tried changing `mainFields` config in https://github.com/farm-fe/farm/blob/main/packages/core/src/config/index.ts#L240 which seems like another approach that would be valid. Are main fields hard coded in rs?

wre232114: The default mainFields are `exports`, `browser`, `module`, `main`, see https://github.com/farm-fe/farm/blob/main/crates/core/src/config/mod.rs#L135.

wre232114: > FWIW, I tried changing `mainFields` config in https://github.com/farm-fe/farm/blob/main/packages/core/src/config/index.ts#L240 which seems like another approach that would be valid. Are main fields hard coded in rs?

It is not hard coded, If you configured the mainFields, the default value will be disabled. And there are many methods to determine the package entry, mainFields is just one of them.

oblador: Hmm, I did some investigation and yes mainField is respected for the entry field, but not for import maps. See this example:
```
{
  "name": "some-package",
  "version": "1.0.0",
  "main": "./main.js",
  "custom": {
    "./implementation.js": "./custom.js"
  }
}
```
It will resolve to `implementation.js` even if `mainFields` have `custom` as first value.

And this:
```
{
  "name": "some-package",
  "version": "1.0.0",
  "main": "./main.js",
  "browser": {
    "./implementation.js": "./browser.js"
  }
}
```
Will resolve to `browser.js` regardless if `browser` is part of `mainFields` or not.

oblador: IMHO it's probably better approach to use a different default `mainFields` config for node targets (if not set) than ignoring `browser` field.

oblador: I cross checked this with how esbuild and webpack works, and they both actually work consistently with how fast behaves, so I think you can ignore my comments.

PR #252, opened 2023-04-23
fix: prefer string type over String

oblador:
**Description:**

Per [TS docs](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#number-string-boolean-symbol-and-object): 

> Don’t ever use the types Number, String, Boolean, Symbol, or Object These types refer to non-primitive boxed objects that are almost never used appropriately in JavaScript code.

**BREAKING CHANGE:**


wre232114 (approved):

wre232114:
Thanks

PR #253, opened 2023-04-23
feat: disable HMR in prod builds

### LINKED ISSUE #249: HMR defaults to true for production builds
### Steps to reproduce

```js
{
  compilation: {
    mode: 'production',
    input: {
      index: './index.ts',
    },
  },
};
```

### What is expected?

Output to not contain development-specific code

Output bundle contains `farm-runtime-hmr-client-plugin` code.

### System Info

```Shell
System:
    OS: macOS 13.3.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 98.73 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 16.18.0 - ~/.asdf/installs/nodejs/16.18.0/bin/node
    Yarn: 1.22.17 - ~/.asdf/shims/yarn
    npm: 8.19.2 - ~/.asdf/plugins/nodejs/shims/npm
    Watchman: 2023.03.27.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 112.0.5615.137
    Firefox: 107.0.1
    Safari: 16.4
  npmPackages:
    @farmfe/core: ^0.6.3 => 0.6.4
```

### Any additional comments?

Should probably default `server.hmr` to `false` for production builds, but it seems it's not the only thing that could be improved.

oblador:
**Description:**

This PR forces `server.hmr` option to false for production builds as there is no dev server in those cases anyways. 

**BREAKING CHANGE:**

Closes #249

### DIFF
--- a/packages/core/src/server/index.ts
+++ b/packages/core/src/server/index.ts
@@ -40,7 +40,7 @@ export class DevServer {
     public logger: Logger,
     options?: UserServerConfig
   ) {
-    this.config = normalizeDevServerOptions(options);
+    this.config = normalizeDevServerOptions(options, "development");
     this._app = new Koa();
 
     // this._app.use(serve(this._dist));
--- a/packages/core/src/config/index.ts
+++ b/packages/core/src/config/index.ts
@@ -27,6 +27,8 @@ export const DEFAULT_CONFIG_NAMES = [
   "farm.config.mjs",
 ];
 
+type CompilationMode = "development" | "production";
+
 /**
  * Normalize user config and transform it to rust compiler compatible config
  * @param config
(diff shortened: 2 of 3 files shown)


packages/core/src/server/index.ts
```diff
@@ -40,7 +40,7 @@ export class DevServer {
     public logger: Logger,
     options?: UserServerConfig
   ) {
-    this.config = normalizeDevServerOptions(options);
+    this.config = normalizeDevServerOptions(options, "development");
```
oblador, line 43: Not really sure why this is needed, the config should have been normalised once already.

wre232114: The compilation config and dev server config are normalized separately. We may merge them as single `normalizeConfig`

oblador: Aha ok. I'm agnostic, we can keep it as is

PR #285, opened 2023-05-04
fix: sfc filename in @farmfe/js-plugin-vue

xGuangsPBD:
**Description:**

A fixed filename is used to generate sfc code, causes same filename in vnode and same component name in Devtools.

**BREAKING CHANGE:**


wre232114:
Thanks!

PR #289, opened 2023-05-05
feat(devserver): Optimize devserver code logic and add proxy configuration

### LINKED ISSUE #286: [Feature]: dev server proxy

add dev server proxy

```ts
    proxy: {
      '/api': {
        target: 'https://music-erkelost.vercel.app/banner',
        changeOrigin: true,
        rewrite: (path: any) => path.replace(/^\/api/, ''),
      },
    },
```

ErKeLost:
**Description:**

Optimize devserver code logic and add proxy configuration

close #286


packages/core/src/server/middlewares/hmr.ts
```diff
@@ -26,3 +28,18 @@ export function hmr(server: DevServer) {
     }
   };
 }
+
+export function hmrPlugin(distance: DevServer) {
```
wre232114, line 32: If I want config a custom middleware, do I need to create a Plugin first?

wre232114: And I think the dev server plugin should not expose to user

wre232114 (approved):

PR #293, opened 2023-05-07
fix: change the path parsing capability for @import in the less plugin

Nirvana-Jie:
**Description:**

Before it.When we use `@import 'xxxx'` in less file. It will look at the root path.I add the the dirname of resolvedPath into the options. This will help `Less` resolve paths better
**BREAKING CHANGE:**
None

None


js-plugins/less/src/index.ts
```diff
@@ -37,6 +38,7 @@ export default function farmLessPlugin(options?: LessPluginOptions): JsPlugin {
                 ? `${await options.additionalData(param.content, this)}`
                 : `${options.additionalData}\n${param.content}`;
           }
+          options.paths.unshift(path.dirname(param.resolvedPath));
```
shulandmimi, line 41: `unshift` -> `concat`?

shulandmimi (changes requested):
https://github.com/farm-fe/farm/blob/1d244481e953fb14a46270e8f661198e80e4869c/js-plugins/less/src/index.ts#L23

`sass` -> `less`

js-plugins/less/src/index.ts
Nirvana-Jie: got it

wre232114 (approved):

shulandmimi (approved):

PR #366, opened 2023-06-05
chore: add format option

Flora025:
**Description:**
add format option in cli


packages/cli/src/types.ts
```diff
@@ -28,6 +28,7 @@ export interface FarmCLIBuildOptions {
   sourcemap?: boolean;
   minify?: boolean;
   treeShaking?: boolean;
+  format?: 'cjs' | 'esm';
```
wre232114, line 31: `format` should be added to `output`, thanks!

wre232114 (changes requested):

packages/cli/src/index.ts
```diff
@@ -94,11 +94,13 @@ cli
         compilation: {
           mode: options.mode,
           watch: options.watch,
-          output: options.outDir
-            ? {
-                path: options.outDir
-              }
-            : undefined,
+          output:
+            options.output.outDir || options.output.format
```
ErKeLost, line 98: Options as a cli command line object is not a property of a compilation object, so there is no output object

packages/cli/src/types.ts
ErKeLost: this is the type of cli It should be right.

PR #404, opened 2023-06-21
style: add the jump link between Chinese and English Readme 

Nirvana-Jie:
**Description:**

add the jump link between Chinese and English Readme and change the js-plugin name

**BREAKING CHANGE:**
None

None

### DIFF
--- a/js-plugins/less/src/options.ts
+++ b/js-plugins/less/src/options.ts
@@ -1,4 +1,6 @@
-export const pluginName = '@farmfe/js-plugin-less';
+import pkg from '../package.json';
+
+export const pluginName = pkg.name;
 
 export function throwError(type: string, error: Error) {
   console.error(`[${pluginName} ${type} Error] ${error}`);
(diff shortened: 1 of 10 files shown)


js-plugins/sass/src/options.ts
```diff
@@ -2,7 +2,7 @@ import { isAbsolute, join } from 'path';
 import type { SassPluginOptions } from './index.js';
 import fs from 'fs';
 
-export const pluginName = 'farm-sass-plugin';
+export const pluginName = '@farmfe/js-plugin-sass';
```
ErKeLost, line 5: Does it feel better to take the name from package json if the name is the same now?

js-plugins/less/src/options.ts
```diff
@@ -1,4 +1,4 @@
-export const pluginName = 'farm-less-plugin';
+export const pluginName = '@farmfe/js-plugin-less';
```
ErKeLost, line 1: Similar to the next question.

js-plugins/sass/src/options.ts
Nirvana-Jie: Yes, I will modify it now

ErKeLost (approved):

Issue #278, opened 2023-04-27
Use `grass` as the replacement of `dart-sass` in rust plugin `sass`


llc1123:
### What problem does this feature solve?

There is [a pure rust implementation of sass](https://github.com/connorskees/grass) which could be a drop-in replacement of `dart-sass`. And it is much faster.

As we are introducing rust plugins in farm, I think it could be better and more elegant getting rid of `dart-sass-embedded`.

### What does the proposed API look like?

Just keep it the same as the current sass plugin.

wre232114:
As I know, sass-embedded is more stable that grass, grass misses many features. And the performance is the same.

We will try grass later, contributions are also welcome, thanks!

llc1123:
> As I know, sass-embedded is more stable that grass

Yes, `sass-embedded` is more stable.

> grass misses many features

No, `grass` is currently having almost every feature of `dart-sass`. Only several edge cases are not covered, which already meets most use cases.

> And the performance is the same.

I didn't do the benchmark, but according to the `README` of the project, `grass` is ~2x faster than `dart-sass`.

On another side, just like `farm`, frontend toolchain projects based on rust need more collaboration to make them widely adopted.

wre232114:
We tried grass last year, and it fails for many projects that we try to migrate, so we switch to dart-sass-embedded, and we also do the benchmark too, the result is close. We will try grass next month, hope that it becomes more reliable!

And yes, we are really glad to make everything in Rust!

PR #451, opened 2023-07-01
feat: Use grass as the replacement of dart-sass in rust plugin sass

NidMo:
**Description:**
Use grass as the replacement of dart-sass in rust plugin sass, delete sass embedded related codes

**BREAKING CHANGE:**
plugin option is currently not supported

close https://github.com/farm-fe/farm/issues/278

### DIFF
--- a/.changeset/modern-pillows-look.md
+++ b/.changeset/modern-pillows-look.md
@@ -1,0 +1,5 @@
+---
+'@farmfe/plugin-sass': major
+---
+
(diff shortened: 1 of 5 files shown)


wre232114:
Please add a changeset for plugin sass.

And Can we rename the current plugin-sass to plugin-sass-embedded, so we can use legacy sass plugin as fallback when there are issues with grass.

wre232114:
I solved the conflicts and opened a new PR #453 ，please just push the new PR if there are more updates.

PR #470, opened 2023-07-12
feat: support add extra watch file

### LINKED ISSUE #457: Support extra watch files for Farm

Some files during compilation are not watched, for example, for a scss file:
```scss
@import "./variable.scss";

body {
  color: $mainColor;
}
```

The rebuild is not triggered when `./variable.css` changed. Cause plugin-sass will merge the two files as one Farm compile module, and only on file will be watched!

Add paths map in both Rust plugins and Js plugins, for example, with a rust plugin:
```rust
impl MyPlugin {
  fn transform(params: FarmPluginTransformHookParams, context: &Arc<CompilationContext>) -> Result<XXX> {
     // ... your operation
     context.addWatchFiles(params.resolvedPath, vec![your_extra_files])
  }
}
```

with a js plugins:
```js
let devServer;

export default {
  name: 'my-plugin',
  configDevServer(server) {
    devServer = server;
  }
  transform: (params, ctx) {
    // ...your operation
    devServer.addWatchFiles(params.resolvedPath, [your_extra_files])
  }
}
```

shulandmimi:
**Description:**

close #457 

**BREAKING CHANGE:**

### DIFF
--- a/crates/core/src/module/watch_graph.rs
+++ b/crates/core/src/module/watch_graph.rs
@@ -69,9 +69,9 @@ impl WatchGraph {
     // v_d     v_e   v_f
     //           \    /
     //            v_g
-    // e change 引起 a、f、c 更新
-    // c change 引起 a
-    // g change 引起 a 更新
+    // e change causes a、f、c update
+    // c change causes a
+    // g change causes a update
     //
     // m  |  root
     // a -> []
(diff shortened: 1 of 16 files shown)


wre232114:
There are conflicts to be resolved

crates/core/src/module/watch_graph.rs
```diff
@@ -0,0 +1,224 @@
...
+      ))
+    })?;
+    //         a                          h               c
+    //       /   \                      /               /
+    //     b      v_c                v_e              v_e
+    //   /       /   \
+    // v_d     v_e   v_f
+    //           \    /
+    //            v_g
+    // e change 引起 a、f、c 更新
```
wre232114, line 72: Use english would be better

wre232114 (changes requested):

crates/core/src/module/watch_graph.rs
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    self
+      .g
+      .update_edge(*from_index, *to_index, EdgeMode::WatchImport);
+
+    Ok(())
+  }
+
+  pub fn resources(&self) -> Vec<&String> {
```
wre232114, line 108: Rename `resources` to `modules`, `resources` is refer to the production files in Farm

crates/core/src/module/watch_graph.rs
```diff
@@ -0,0 +1,220 @@
...
+    let mut res = HashSet::new();
+
+    for node in self.g.edge_indices() {
+      match self.g.edge_weight(node).unwrap() {
```
wre232114, line 112: Fix clippy issue here

wre232114 (changes requested):

wre232114 (approved):

Issue #494, opened 2023-07-28
[Bug] farm incorrect resolve dependency


NidMo:
### Steps to reproduce

1. cd example/solid
2.  add counter.tsx
```tsx
import { createSignal } from "solid-js";
import { customElement } from "solid-element";

customElement("my-counter", () => {
    const [count, setCount] = createSignal(0);
    return (
        <>
            <div>
                <button onClick={() => setCount(count() - 1)}>-</button>
                <span>{count()}</span>
                <button onClick={() => setCount(count() + 1)}>+</button>
            </div>
        </>
    );
});
```

3. change index.html
```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1" />
    <meta name="theme-color" content="#000000" />
    <link rel="shortcut icon" type="image/ico" href="/src/assets/favicon.ico" />
    <title>Solid App</title>
  </head>
  <body>
    <noscript>You need to enable JavaScript to run this app.</noscript>
    <div id="root">
      <my-counter></my-counter>
    </div>

    <script src="./src/counter.tsx" type="module"></script>
  </body>
</html>
```

### What is expected?

import { createSignal } from "solid-js";

should resolve node_modules/solid-js/dist/solid.js

### What is actually happening?

resolve node_modules/solid-js/dist/server.js

### System Info

```Shell
System:
    OS: macOS 13.4.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 74.02 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 16.14.0 - ~/.nvm/versions/node/v16.14.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.14.0/bin/yarn
    npm: 9.1.2 - ~/.nvm/versions/node/v16.14.0/bin/npm
    pnpm: 8.6.2 - ~/.nvm/versions/node/v16.14.0/bin/pnpm
    Watchman: 2023.04.10.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 115.0.5790.114
    Safari: 16.5.2
```

PR #497, opened 2023-07-29
fix(js-plugins/solid): HMR assertion

wjq990112:
**Description:**

Fix the `needHmr` assertion in plugin.

**BREAKING CHANGE:**

None.

Part of #494.


js-plugins/solid/rollup.config.js
```diff
@@ -0,0 +1,52 @@
+import cjs from '@rollup/plugin-commonjs';
```
wre232114, line 1: Use farm to bundle plugin solid

wre232114 (changes requested):

wre232114 (approved):

PR #521, opened 2023-08-23
feat: enable port auto-increment to prevent port conflict (#519)

### LINKED ISSUE #288: [Feature]: dev server ws port conflict and injection

Solve the problem of port conflict and injection under multiserver

port conflicts need to be prevented when starting multiple servers at the same time

It is recommended that websocket and devserver use the same port for output

devserver and websocket prot conflict
devserve and ws use same port cause complate code
Injecting environment variables to listen for client ws hmr updates

none

Flora025:
**Description:**

An initial attempt to resolve port conflict when starting multiple servers at the same time (#519)
Temporarily achieved by running an HTTP server to check port availability before normalizing the configs

Current behavior: dev server will start on port 9000 by default, and the following servers on port 9001, 9002... The hmr port will also ++ at the same time.

close #288 
**BREAKING CHANGE:**

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -20,7 +20,7 @@ import {
   UserConfig
 } from './config/index.js';
 import { DefaultLogger } from './utils/logger.js';
-import { DevServer } from './server/index.js';
+import { DevServer, resolvePortConflict } from './server/index.js';
 import { FileWatcher } from './watcher/index.js';
 import { Config } from '../binding/index.js';
 import { compilerHandler } from './utils/build.js';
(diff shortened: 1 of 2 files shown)


ErKeLost:
Two ways that can be adopted at present: 
1. Start a `http server` to determine port access before the hmr plugin hook is registered. If there is a `port conflict`, the `port + +`
2. We can get part of the configuration file and start devserver first to determine `port conflicts` (instead of creating a http to determine), and then compile (this method requires an attempt to load the browser, but the compilation is not complete, or we can not let the user feel that we have started the service and wait for the compilation to complete before entering the result in `node`).

Issue #463, opened 2023-07-06
Support more `partial bundling` options


wre232114:
### What problem does this feature solve?

Partial Bundling does not support merge modules based on the module size for now, and It only bundles the modules based on the module group and configured module buckets.

we need to support following options to make partial bundling more powerful:
* `minResourceSize`: resource size less than this value will be merged.
* `maxConcurrentRequests`: max concurrent requests when loading initial resources or dynamic resources.

see https://github.com/farm-fe/rfcs/pull/9 for details.

### What does the proposed API look like?

see above.

PR #532, opened 2023-08-30
Refactor/partial bundling

shulandmimi:
**Description:**

**BREAKING CHANGE:**

#463


wre232114:
I think following designs can be optimized:
1. We have already designed `mutable/immutable modules` so default configuration like `vendors` is not needed any more.
2. We do not `split` but do `merge`, see [RFC-003](https://github.com/farm-fe/rfcs/pull/9)

Issue #538, opened 2023-09-07
hmr's websocket service is flawed


chovrio:
### Steps to reproduce

Repeat the start script

### What is expected?

When the port is occupied, the port number is automatically incremented by 1.

### What is actually happening?

hmr's websocket service does not have an automatic port number +1 if the port is occupied

### System Info

```Shell
System:
    OS: Windows 10 10.0.19045
    CPU: (8) x64 Intel(R) Xeon(R) CPU E3-1545M v5 @ 2.90GHz
    Memory: 2.72 GB / 15.88 GB
  Binaries:
    Node: 18.13.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files\nodejs\yarn.CMD
    npm: 8.19.3 - C:\Program Files\nodejs\npm.CMD
    pnpm: 8.5.1 - C:\ProgramData\chocolatey\bin\pnpm.EXE
  Browsers:
    Chrome: bookmarks.html
initial_preferences
    Edge: Spartan (44.19041.1266.0), Chromium (116.0.1938.69)
    Internet Explorer: 11.0.19041.1566
  npmPackages:
    @farmfe/core: 0.12.1 => 0.12.1
```

### Any additional comments?

[
![Snipaste_2023-09-07_21-48-33](https://github.com/farm-fe/farm/assets/92280658/20d0aba9-c353-48b8-96e9-765977a39493)
](url)

PR #540, opened 2023-09-07
Solve the problem of multiple farm project start port conflicts

chovrio:
#538 
**Description:Multiple farm projects run port conflicts at  same time**

**BREAKING CHANGE:No, only one function has been modified.**

**Related issue (if exists):Solve the problem of multiple farm project start port conflicts**
farm1
farm2
farm3


wre232114:
thanks

Issue #545, opened 2023-09-08
[Feature] reuse http server when hmr is not configured


FairyScript:
### What problem does this feature solve?

In my case, I use farm under a reverse proxy. that means if all of Farm's functionality can be accessed with just one port, it will greatly reduce the difficulty of configuration.

### What does the proposed API look like?

in `config.server`:

```js
{
  port: 9000,
  hmr: {
    port: 9000,// if hmr port is same as server.port, reuse the server
    path: '/__hmr'// default ws path.
  }
}
```

PR #546, opened 2023-09-08
feat: reuse http server when hmr is not configured

FairyScript:
**Description:**

when `hmr.host` & `hmr.port` is same as dev server, reuse the dev server to handle hmr request.

**BREAKING CHANGE:**

if reuse_mode, ws server need a unique path, so add a new property  `config.server.hmr.path`. the default value is '__hmr'

```ts
//server.hmr
{
path?:string; //default: '__hmr'
}
```

#545


ErKeLost:
thanks

PR #573, opened 2023-09-20
feat: generate stage add record

NidMo:
**Description:**
1. generate stage add record
2. remove temporary macro hook_first_with_callback
3. remove temporary macro hook_serial_with_callback
4. config add record switch
5. export some function to node env:
get_resource_pot_records_by_id

**Usage:**

```ts
// farm.config.ts
// set record as true , open record switch
export default {
  compilation: {
    ...,
    record: true
  },
};
```
http://localhost:9000/_resource_pot_records?id=xxx
return resource pot records

**BREAKING CHANGE:**

no


crates/core/src/record/mod.rs
```diff
@@ -85,6 +95,8 @@ impl RecordManager {
 
   pub fn get_transform_records_by_id(&self, id: &str) -> Vec<TransformRecord> {
     let transform_map = self.transform_map.read().unwrap();
+    let a = self.get_resource_pot_records_by_id(id);
+    println!("get_resource_pot_records_by_id:{:?}", a);
```
wre232114, line 99: Remove `println`

NidMo: done

wre232114 (approved):

PR #661, opened 2023-10-26
feat: js plugin record viewer

NidMo:
**Description:**
1.  add js plugin record viewer
2.  export modules function to node env
3.  change record api path

**BREAKING CHANGE:**

Before:

http://localhost:9000/_resolve_records
return resolve ids
http://localhost:9000/_transform_records?id=xxx
return module transom records
http://localhost:9000/_process_records?id=xxx
return module process records
http://localhost:9000/_analyze_deps_records?id=xxx
return module analyze deps records

After:

http://localhost:9000/__record/modules
return modules
http://localhost:9000/__record/resolve
return resolve ids
http://localhost:9000/__record/transform?id=xxx
return module transom records
http://localhost:9000/__record/process?id=xxx
return module process records
http://localhost:9000/__record/analyze_deps?id=xxx
return module analyze deps records


packages/core/src/server/middlewares/records.ts
ErKeLost, line 1: Maybe we can optimize this code and use the match syntax.

NidMo: optimization will be carried out in the future

PR #696, opened 2023-11-01
refactor: record  add hmr flag

NidMo:
**Description:**
- resolve record add is hmr filed
- transform record add is hmr filed
- record manager add trigger filed

**BREAKING CHANGE:**
No

### DIFF
--- a/crates/core/src/record/mod.rs
+++ b/crates/core/src/record/mod.rs
@@ -8,10 +8,8 @@ use crate::{
 };
 
 #[derive(Debug, Clone)]
-pub enum Stage {
-  Init,
-  Build,
-  Generate,
+pub enum Trigger {
+  Compiler,
   Update,
 }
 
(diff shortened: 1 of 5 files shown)


crates/core/src/record/mod.rs
```diff
@@ -1,8 +1,19 @@
...
 
-use crate::{module::{ModuleId, ModuleType}, plugin::{PluginAnalyzeDepsHookResultEntry, ResolveKind}};
+use crate::{
+  module::{ModuleId, ModuleType},
+  plugin::PluginAnalyzeDepsHookResultEntry,
+};
+
+#[derive(Debug, Clone)]
+pub enum Stage {
+  Init,
```
wre232114, line 12: what does `Init` mean

wre232114: but we only have 2 stages: build and generate, I think we can rename it to avoid name conflict

NidMo: No use stage to name, use something like status, mode, phase ？Currently only used to identify whether it is an HMR, but it may be expanded to a stage in the future

NidMo: rename filed stage as trigger

wre232114 (approved):

Issue #729, opened 2023-11-09
✨ [Feature]: Supoort import.meta.url and import.meta.env


ErKeLost:
### What problem does this feature solve?

- [ ] support import.meta.url. the value of `import.meta.url` is the url of current production resource. for example: `/index-ec34.js`
- [ ] support import.meta.env, align to https://vitejs.dev/guide/env-and-mode.html#env-variables

### What does the proposed API look like?

-

PR #738, opened 2023-11-14
feat: Supoort import.meta.url and import.meta.env

NidMo:
**Description:**
1. add runtime plugin import-meta
2. render_resource_pot_modules inject resource pot name

> ### env

- import.meta.env.mode: {string} the mode the app is running in.
- import.meta.env.prod: {boolean} whether the app is running in production.
- import.meta.env.dev: {boolean} whether the app is running in development 

> ### url
- import.meta.url: {string} the path of the resource to which the module belongs

**BREAKING CHANGE:**
no

close https://github.com/farm-fe/farm/issues/729


packages/core/src/config/index.ts
```diff
@@ -233,6 +236,19 @@ export async function normalizeUserCompilationConfig(
...
   }
 
+  if (
+    config.output.targetEnv !== 'node' &&
+    isArray(config.runtime.plugins) &&
+    !config.runtime.plugins.includes(ImportMetaPluginPath)
+  ) {
+    config.runtime.plugins.push(ImportMetaPluginPath);
+    config.define.FARM_PORT = String(normalizedDevServerConfig.port);
+    config.define.FARM_HOST =
```
ErKeLost, line 246: can't use hostname directly in this place. The default hostname is localhost. If he configured host true, then it would be wrong.

can use window.location.hostname directly at runtime

packages/runtime/src/module-system.ts
```diff
@@ -184,6 +198,19 @@ export class ModuleSystem {
     }
   }
 
+  getModuleUrl(moduleId: string): string {
+    for (const [resource, modules] of this.resources) {
+      if (modules.includes(moduleId)) {
+        const publicPath = this.publicPaths[0];
+        // @ts-ignore
+        const url = `${window.FARM_HOST}:${window.FARM_PORT}${
```
ErKeLost, line 206: refer to the previous

NidMo: like const url = `${location.host}${publicPath === '/' ? '' : publicPath}/${resource}`;  ？

ErKeLost: can do this for a while, because it's uncertain.

packages/runtime-plugin-import-meta/src/index.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export default <FarmRuntimePlugin>{
+  name: 'farm-runtime-import-meta',
+  _moduleSystem: {} as ModuleSystem,
+  bootstrap(system: ModuleSystem) {
+    this._moduleSystem = system;
+  },
+  moduleCreated(module) {
+    module.meta.env = {
+      ...process.env,
```
ErKeLost, line 12: Does this process.env contain all the environment variables of node? if so, we should not use full process.env.

NidMo: So should we still limit the variables in import.meta.env? At first, I didn't want to inherit process.env, but if developers customize some env, import.meta.env won't be available.

ErKeLost (approved):

PR #748, opened 2023-11-20
fix: location is not define and bump version

ErKeLost:
**Description:**

**BREAKING CHANGE:**


packages/runtime-plugin-import-meta/src/index.ts
```diff
@@ -8,19 +10,22 @@ export default <FarmRuntimePlugin>{
...
     const publicPath = this._moduleSystem.publicPaths[0];
 
-    if (this._moduleSystem.targetEnv === 'node') {
-      module.meta.url = module.resource_pot;
-    } else {
-      const url = `${location.host}${publicPath === '/' ? '' : publicPath}/${
+    if (__global_this__.location) {
+      const url = `${__global_this__.location.protocol}//${
+        __global_this__.location.host
+      }${publicPath.endsWith('/') ? publicPath.slice(0, -1) : publicPath}/${
```
wre232114, line 23: publicPath may be undefined

packages/runtime/src/module-system.ts
```diff
@@ -202,10 +204,17 @@ export class ModuleSystem {
 
   getModuleUrl(moduleId: string): string {
     const publicPath = this.publicPaths[0];
-    const url = `${location.host}${publicPath === '/' ? '' : publicPath}/${
-      this.modules[moduleId].__farm_resource_pot__
-    }`;
-    return url;
+
```
wre232114, line 207: this logic may be shared between plugin-import-meta and runtime

wre232114 (approved):

PR #849, opened 2023-12-18
feat: vue & solid ssr examples

shulandmimi:
**Description:**

- vue & solid ssr examples
- plugin-resolve package export condition order
- vue & solid plugin ssr option

**BREAKING CHANGE:**

### DIFF
--- a/examples/solid-ssr/farm.config.mjs
+++ b/examples/solid-ssr/farm.config.mjs
@@ -1,4 +1,4 @@
-import solid from '@farmfe/js-plugin-solid';
+import solid from 'vite-plugin-solid';
 import { generateHydrationScript } from 'solid-js/web';
 
 /**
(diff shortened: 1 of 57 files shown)


examples/vue-ssr/farm.config.mjs
```diff
@@ -0,0 +1,52 @@
+import vue from '@farmfe/js-plugin-vue';
```
wre232114, line 1: we should use @vitejs/plugin-vue, `@farmfe/js-plugin-vue` is deprecated

examples/solid-ssr/farm.config.mjs
```diff
@@ -0,0 +1,62 @@
+import solid from '@farmfe/js-plugin-solid';
```
wre232114, line 1: we should use vite-plugin-solid here too.

wre232114 (approved):

PR #854, opened 2023-12-25
feat: add overlay window and optimize socket client interaction code

ErKeLost:
**Description:**

**BREAKING CHANGE:**


packages/core/src/server/hmr-engine.ts
```diff
@@ -139,6 +140,19 @@ export class HmrEngine {
...
       } catch (e) {
+        // eslint-disable-next-line no-control-regex
+        const serialization = e.message.replace(/\x1b\[[0-9;]*m/g, "");
+        const errorStr = `${JSON.stringify({
+          message: serialization
+        })}`;
+        this._devServer.ws.clients.forEach((client: WebSocketClient) => {
+          client.rawSend(`
+            {
+              type: 'farm-error',
```
wre232114, line 151: we can reuse `error` of vite here, make the structure the same as vite, it can be a string or a object

Issue #856, opened 2023-12-27
[Bug Report]: process is not define


ErKeLost:
### Steps to reproduce

`pnpm create farm` start

### What is expected?

normal

### What is actually happening?

![image](https://github.com/farm-fe/farm/assets/66500121/457a95f4-2521-4aa6-a3f4-a26e2d5c1212)

![image](https://github.com/farm-fe/farm/assets/66500121/80231837-cd82-490a-9143-c4ac41c253da)

### System Info

```Shell
macos
```

PR #861, opened 2024-01-02
fix: process is not define

ErKeLost:
**Description:**

close: #856 

**BREAKING CHANGE:**


superlbr:
process is not define, but import.meta

PR #866, opened 2024-01-06
fix(js-plugins/dts): change options to optional

wjq990112:
**Description:**

Change options to optional.

**BREAKING CHANGE:**

None.

None.


ErKeLost:
The version will be released uniformly in the next big version of @ farmfe/core

PR #910, opened 2024-01-23
chore: update other template `lit` `vanilla` 

ErKeLost:
**Description:**

add other template 

optimize create farm

**BREAKING CHANGE:**


configs/farm-js-plugin.base.config.mjs
```diff
@@ -15,11 +15,6 @@ export function createFarmJsPluginBuildConfig(plugins, options = {}) {
         targetEnv: 'node',
         format
       },
-      external: [
```
wre232114, line 18: why remove external

Issue #964, opened 2024-02-06
Can't find peerDependencies when using the yarn as package manager.


motea927:
### Steps to reproduce

```shell
yarn create farm my-vue-app --template vue
cd my-vue-app
yarn install
yarn start
```

### What is expected?

worked

### What is actually happening?

get missing dependencies error

<img width="1241" alt="image" src="https://github.com/farm-fe/farm/assets/25958590/3c9f683a-7ed2-405f-9955-0e82ce8bbe62">

### System Info

```Shell
System:
    OS: macOS 13.6.1
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 238.37 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.10.0 - ~/.nvm/versions/node/v20.10.0/bin/node
    Yarn: 1.22.15 - /usr/local/bin/yarn
    npm: 10.2.3 - ~/.nvm/versions/node/v20.10.0/bin/npm
    pnpm: 8.15.1 - ~/.nvm/versions/node/v20.10.0/bin/pnpm
    bun: 1.0.15 - /usr/local/bin/bun
    Watchman: 2023.08.14.00 - /usr/local/bin/watchman
  Browsers:
    Brave Browser: 119.1.60.118
    Chrome: 121.0.6167.139
    Safari: 17.1
    Safari Technology Preview: 17.4
  npmPackages:
    @farmfe/core: ^0.16.3 => 0.16.5
```

PR #965, opened 2024-02-06
fix: resolve cannot find package 'vite' when using the yarn as packag…

motea927:
**Description:**

Resolve cannot find package 'vite' when using the yarn as package manager, adding `vite` into `devDependencies`.

Or we can add [.npmrc](https://github.com/nuxt/nuxt/blob/main/.npmrc) and enable `shamefully-hoist` to support more pacakge manager

**BREAKING CHANGE:**

#964

### DIFF
--- a/packages/create-farm/templates/vue/package.json
+++ b/packages/create-farm/templates/vue/package.json
@@ -8,7 +8,8 @@
     "@farmfe/cli": "^0.6.1",
     "@farmfe/core": "^0.16.5",
     "@vitejs/plugin-vue": "^4.5.0",
-    "core-js": "^3.30.1"
+    "core-js": "^3.30.1",
+    "vite": "^5.0.12"
   },
   "scripts": {
     "dev": "farm start",


wre232114:
thanks

ErKeLost:
hi @motea927 Directly treating `vite` as `devDependency` may not be the best choice. This may modify all template and we are just use `plugins` for `vite`. At present, `yarn` does not support automatic installation of peers dependency. In the current method, we can determine whether users are currently using yarn to run in create-farm. If we use yarn to modify the package.json of the current template to add vite, do you have a better way? I saw you talking about `shamefully-hoist`, but this is supposed to only work on `pnpm`.

ErKeLost:
@motea927  I tried npm and pnpm, and they both install peers dependencies automatically, and I think it's a good idea for you to add script, or by judging if use yarn installation. like `"postinstall": "npx check-peer-dependencies --yarn --install"` Dynamically add a script to the scripts inserted into the package json.

If you want to try. Welcome pr 🎉

PR #985, opened 2024-02-18
feat: add postinstall script for peer-gear when Yarn is installed

motea927:
**Description:**

This pull request addresses the discussions in #965 by dynamically adding a script to the `package.json` when using Yarn. The purpose is to ensure that Yarn installs peer dependencies in a manner consistent with `npm` and `pnpm`.

Previously, we used the [check-peer-dependencies](https://github.com/christopherthielen/check-peer-dependencies) package for this purpose. While it is a reliable tool, it has a drawback – it installs `pre-release` (possibly `beta`) versions. To overcome this limitation, I've forked the repository and made modifications. The forked version now defaults to disabling `includePrerelease`, and I've included additional unit tests to enhance robustness.

Through thorough testing across various templates (including `Vanilla`, `React`, `Vue`, `Solid`, `Svelte`, `Lit`), this change has been validated, and all templates function as expected. This adjustment aims to improve the Developer Experience (DX) for those using Yarn as their package manager.

**BREAKING CHANGE:**

#964

### DIFF
--- a/packages/create-farm/index.ts
+++ b/packages/create-farm/index.ts
@@ -160,6 +160,11 @@ function writePackageJson(dest: string, options: IResultType) {
 
   pkg.name = options.projectName;
 
+  if (isYarnInstalled) {
+    pkg.scripts = pkg.scripts ?? {};
+    pkg.scripts.postinstall = 'npx --yes peer-gear --install';
+  }
+
   const packageJsonPath = path.join(dest, 'package.json');
   const { name, ...rest } = pkg;
   const sortedPackageJson = { name, ...rest };


packages/create-farm/index.ts
```diff
@@ -160,6 +160,11 @@ function writePackageJson(dest: string, options: IResultType) {
 
   pkg.name = options.projectName;
 
+  if (isYarnInstalled) {
```
wre232114, line 163: `options.packageManager` should be used here, make sure this logic executed only when the user chooses `yarn` as pacakge manager. And `pacakgeManager` should be passed as arguments in https://github.com/farm-fe/farm/pull/985/files#diff-53f16bc3fa79da2028e77129b2f7ad141dac3fb81fcd74a7e1e44f8cf5aff945R129

wre232114 (changes requested):

packages/create-farm/index.ts
motea927: When `options.packageManager` is directly used, it doesn't accurately identify `yarn` in the `yarn create farm` command.

To resolve this, I've used the same strategy as `installationDeps` by utilizing `currentPkgManager`.

This ensures alignment with the prompt and also properly recognizes `yarn`. 

Could you please share your thoughts on this?

wre232114: Thanks, that's exactly what i think

wre232114 (approved):

PR #992, opened 2024-02-27
feat: add star history

liangchaofei:
add star history

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,7 +222,7 @@ Thanks to:
 
 - The [Vite](https://github.com/vitejs/vite) project created by [Evan You](https://github.com/yyx990803), which inspired Farm's compatibility design of ecosystem.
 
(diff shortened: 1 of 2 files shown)


wre232114:
I think we can add a new `start_history.md` file that shows our star history.

liangchaofei:
> I think we can add a new `start_history.md` file that shows our star history.

See what other people suggest?

ErKeLost:
hi @liangchaofei . Thank you for your contribution.
My personal view is that farm is about to release `1.0`, our readme to modify the directory structure should be shortened, should mainly introduce our features and features as the main point, such as our benchmark and ecological introduction need to be deleted, if you like, you can create a new `start_history.md` record, or we directly display on the official document.  Personally, I think farm's readme may not need to show star history.

PR #1025, opened 2024-03-14
feat: minify modules instead of resource pots

### LINKED ISSUE #951: Refactor plugin_minify to optimize minify performance

Currently minification is slower than it could be, we need to refactor the implementation to make minify faster

try minify by modules instead of resource pots

shulandmimi:
**Description:**

- minify modules instead of resource pots
- sourcemap field deserialize

**BREAKING CHANGE:**

close #951


crates/plugin_runtime/src/render_resource_pot/mod.rs
```diff
@@ -14,17 +16,19 @@ use farmfe_core::{
   error::{CompilationError, Result},
   module::{module_graph::ModuleGraph, ModuleId, ModuleSystem},
   parking_lot::Mutex,
-  rayon::prelude::*,
+  rayon::iter::{IntoParallelIterator, ParallelIterator},
   resource::resource_pot::{RenderedModule, ResourcePot},
-  swc_common::{comments::SingleThreadedComments, Mark},
+  swc_common::{comments::SingleThreadedComments, Mark}, // swc_ecma_ast::Function
 };
+use farmfe_plugin_minify::minify_js_module;
```
wre232114, line 23: plugins should not be depend on each other, extract minify_js_module to farmfe_toolkit

wre232114 (changes requested):

crates/core/src/config/mod.rs
```diff
@@ -41,7 +49,7 @@ pub struct Config {
   pub lazy_compilation: bool,
   pub core_lib_path: Option<String>,
   pub tree_shaking: bool,
-  pub minify: Box<BoolOrObj<serde_json::Value>>,
```
wre232114, line 44: we should use `serde_json::Value` here, or it may break Rust plugins

rust-plugins/react/Cargo.toml
```diff
@@ -11,8 +11,8 @@ farmfe_core = { version = "*", path = "../../crates/core" }
 farmfe_macro_plugin = { version = "*", path = "../../crates/macro_plugin" }
 farmfe_toolkit_plugin_types = { version = "*", path = "../../crates/toolkit_plugin_types" }
 serde = { version = "1.0", features = ["derive"] }
+farmfe_toolkit = { path = "../../crates/toolkit" }
```
wre232114, line 14: plugin should not relies on `farmfe_toolkit` to avoid swc status conflicts, use `farmfe_core` instead

wre232114 (changes requested):

wre232114 (approved):

Issue #911, opened 2024-01-24
support farm resolver for postcss-import


wre232114:
### What problem does this feature solve?

for postcss plugin like `postcss-import`, it needs context info of `@farmfe/js-plugins-postcss`, e.g. `context.resolve` to resolve farm alias.

### What does the proposed API look like?

inline `postcss-import` into `@farmfe/js-plugins-postcss` like vite:
<img width="677" alt="image" src="https://github.com/farm-fe/farm/assets/8372439/394aa883-0625-4f1a-9dcb-eeba197c2407">

 disabled by default.

```ts
// farm.config.ts
export default defineConfig({
    plugins: [postcss({
        internalPlugins: {
            'postcss-import': true,
        }
    })]
});
```

Maidang1:
I will attempt it, If I have time this week.

wre232114:
Thanks!

PR #1050, opened 2024-03-19
feat: support postcss-import (#911)

Maidang1:
**Description:**

import postcss-import plugin to support inline css

**BREAKING CHANGE:**
no

#911


js-plugins/postcss/src/index.ts
```diff
@@ -53,6 +68,49 @@ export default function farmPostcssPlugin(
...
+                  const publicFile = await checkPublicFile(id, userConfig);
+                  if (publicFile) {
+                    return publicFile;
+                  }
+
+                  try {
+                    const resolvedInfo = await context.resolve(
+                      {
+                        kind: 'import',
+                        importer: param.moduleId,
```
wre232114, line 87: I think the importer should be `baseDir` or `baseDir/xxx` here. For import  hierarchy that more that 2, it may have issues. For example, `@import '../index.css'`, and in `../index.css` it has `@import './stylecss'`. Then `@import './stylecss'` should be resolved from `../index.css` instead of `param.moduleId`

js-plugins/postcss/src/index.ts
```diff
@@ -53,6 +68,49 @@ export default function farmPostcssPlugin(
       async executor(param, context) {
         try {
           const sourceMapEnabled = context.sourceMapEnabled(param.moduleId);
+          const enablePostcssImport =
+            options.internalPlugins?.postcssImport || true;
```
wre232114, line 72: postcssImport should be set to false

wre232114 (approved):

PR #1079, opened 2024-03-27
Support modulePreload when generating html

aniketkumar7:
Sure, here's the updated description with "I" instead of "We":

This pull request addresses the issue of prefetching resources for dynamic imports in Farm. The current implementation of prefetching in Farm only handles static resources, but dynamic imports require a more sophisticated approach to ensure that resources are prefetched at the correct time.

To solve this issue, I introduced a new `DynamicPrefetchResource` struct that represents a resource that should be prefetched when its corresponding module is loaded. I also added a new `dynamic_prefetch` field to the `ResourcesInjectorOptions` struct that is a vector of `DynamicPrefetchResource` objects.

In the `ResourcesInjector` class, I added a new `inject_dynamic_prefetch` method that injects a `<link>` element with `rel="prefetch"` for each dynamic prefetch resource. I also added an `onload` event listener to the `<link>` element that will prefetch the dynamic imports of the corresponding module when the `<link>` element has finished loading. The `onload` event listener uses the `getDynamicModuleResourcesMap` method from the `FARM_MODULE_SYSTEM` to get the dynamic imports of the corresponding module.

Finally, I modified the `visit_mut_element` method to call the `inject_dynamic_prefetch` method in addition to the `inject_preload_and_prefetch` method when processing the `<head>` element.

With these changes, Farm can now prefetch resources for dynamic imports at the correct time, improving the performance of applications that rely on dynamic imports.

### DIFF
--- a/crates/plugin_html/src/resources_injector.rs
+++ b/crates/plugin_html/src/resources_injector.rs
@@ -1,4 +1,5 @@
-use std::{collections::HashMap, sync::Arc};
+use std::{collections::HashMap, fmt::Display, sync::Arc};
+use url::Url;
 
 use farmfe_core::{
   config::{Mode, FARM_MODULE_SYSTEM},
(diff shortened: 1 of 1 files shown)


wre232114:
Could you add a example `html-preload` with a `e2e.test.ts` under `examples` to test it and make sure it works

Issue #1136, opened 2024-04-07
Too strict html parsing.


houd1ni:
### Steps to reproduce

Have a html that has templating stuff in it for ssr:

```html

<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8" />
    {head}
    {styles}
  </head>
  <body>
    <div id="app">{ssr}</div>
    <script type="module" src="./src/main.ts"></script>
  </body>
</html>

```

### What is expected?

Those are considered as text and ignored like vite does (I'm moving this project from vite to farm as it looks more concise, managed and progressive).

### What is actually happening?

The parser got scrambled and throws:

![image](https://github.com/farm-fe/farm/assets/7501201/ea0e9bd4-77ed-4e57-ae34-1ec264ab5c25)

### System Info

```Shell
Win10x64 Pro latest, Intel 14600kf, 32Gb ram, almost empty.
```

### Any additional comments?

If you have any preffered workaround for this case, please tell me.

wre232114:
Thanks for reporting issues. I'll make it less strict today and release `@farmfe/core@1.0.15`. For earlier versions, you can use html tags as placeholders like:

```html
<div id="root">
  <div>app-html-to-replace</div>
</div>
```

wre232114:
please try `@farmfe/core@1.0.15`

houd1ni:
```
 > npm i @farmfe/core@1.0.15


added 1 package, removed 1 package, changed 1 package, and audited 806 packages in 10s

190 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

```

```
> npm run d-dev

> menu-front@0.0.1 dasha-dev
> farm start ./d-c ./.../farm.config.ts

[ Farm ] Using config file at C:\Users\...farm.config.ts
[ Farm ] Port 9000 is in use, trying another one...
[ Farm ] Port 9001 is in use, trying another one...
[ building ] ⠂ transform (15) ../node_modules/@farmfe/runtime-plugin-hmr/src/overlay.ts.farm-runtime                                         [ Farm ] Failed to start the server: Error: Parse `index.html` failed.
 Error: Parse `index.html` failed.
 Error:
  × Stray end tag "head"
   ╭─[index.html:4:1]
 4 │     <meta charset="utf-8" />
 5 │     {head}
 6 │     {styles}
 7 │   </head>
   ·   ───────
 8 │   <body>
 9 │     <div id="app">{ssr}</div>
[...]

PR #1175, opened 2024-04-12
fix: lazy compile error when mixed import and dynamic import

wre232114:
**Description:**

**BREAKING CHANGE:**


houd1ni:
https://github.com/farm-fe/farm/issues/1136 is totally relative.

Issue #1086, opened 2024-03-28
disable polyfill when entry is not html


wre232114:
### What problem does this feature solve?

do not inject polyfill by default when bundling libraries

### What does the proposed API look like?

non

CCherry07:
Let me try  ✅

PR #1183, opened 2024-04-14
feat: disable polyfill when entry is not html

CCherry07:
**Description:**

**BREAKING CHANGE:**

#1086


.changeset/cool-goats-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@farmfe/core': major
```
wre232114, line 2: we should not bump @farmfe/core as major. minor would be enough

wre232114 (changes requested):

wre232114 (approved):

wre232114 (approved):

PR #1221, opened 2024-04-20
feat: disable polyfill when entry is not html

CCherry07:
**Description:**

**BREAKING CHANGE:**


ErKeLost:
Thanks, I think our `create-farm` needs to be as simple as possible, the `ssr` template we can provide to users right out of the box, but we as a `framework template`, such as `css selector` should not be necessary, we should provide users with a simple version of template and should not build too many other options. I think hints such as `css selectors` should be provided by `higher-level cli`, and we should only need to provide a `ssr template` at present.

CCherry07:
> Thanks, I think our `create-farm` needs to be as simple as possible, the `ssr` template we can provide to users right out of the box, but we as a `framework template`, such as `css selector` should not be necessary, we should provide users with a simple version of template and should not build too many other options. I think hints such as `css selectors` should be provided by `higher-level cli`, and we should only need to provide a `ssr template` at present.
> 
> It is also unnecessary to keep the user selective and concise distinction between `js` and `ts`. `Farm` natively supports the compilation of `ts` and `js`, and there is almost no difference in the code except for the file suffix.
> 
> So we should come to the following conclusion
> 
> * Add `ssr` template
> * Remove the prompts from the `css selector` and `js` template， keeping our default ts mode

I converged the css selector and js template to argv-e

ErKeLost:
I'll test these templates later today. Thank you.

ErKeLost:
After I merge the template of tauri, I will merge this pr.

ErKeLost:
hi @CCherry07
   I've taken a closer look at your code, and we need to consider pulling out the entire `template` right now, because we need to maintain templates for other frameworks such as tauri,electron, and I think we need to consider some of the underlying configurations, which will increase the maintainability of `create-farm`, because we're not sure what will happen in the future and what users will do with `create-farm`. I will keep the pr for the time being and make a decision after pulling out the whole `template`. I think we still need the `template` of `ssr`.

Issue #1224, opened 2024-04-21
support receive Promise plugin 


shulandmimi:
### What problem does this feature solve?

```ts
// farm.config.ts
const loadAsyncJsPlugin = async () => ({ ... });
const loadAsyncArrJsPlugin = async () => [{ ... }, { ... }]
const loadAsyncRustPlugin async () => '@farmfe/rust-plugin';
const loadAsyncArrRustPlugin async () => [['@farmfe/rust-plugin', options], ...];

export default {
	plugins: [loadAsyncJsPlugin(), loadAsyncArrJsPlugin(), loadAsyncRustPlugin(), loadAsyncArrRustPlugin()]
}
```

### What does the proposed API look like?

modify

- [packages/core/src/plugin/index.ts](https://github.com/farm-fe/farm/blob/71b6bab71741b9289b6f5b6c2b330b89c41eaeea/packages/core/src/plugin/index.ts#L12-L57)
- [packages/core/src/plugin/index.ts](https://github.com/farm-fe/farm/blob/71b6bab71741b9289b6f5b6c2b330b89c41eaeea/packages/core/src/plugin/index.ts#L60-L74)

combine these two function

DeepeshKalura:
@shulandmimi  I raised the PR tyres to resolve your issue could please look at it and let me know any improvements needed 

https://github.com/farm-fe/farm/pull/1255

shulandmimi:
@DeepeshKalura Thank you very much for your PR, but I see it has not been updated recently, do you have the energy to achieve the rest of it? I can assign this issue to you if you like.

DeepeshKalura:
@shulandmimi Yeah I would love to work in this issue. The reason I not been updating recently because I was getting a bug while working on this issue `Error: Cannot find module '019171a2'` while running with `pnpm bootstrap`.  Debugging this issue cause lot's of time.

I got a suggestion to re-clone the project and restart the work from discord farm community. Working on that part.

ErKeLost:
> @shulandmimi Yeah I would love to work in this issue. The reason I not been updating recently because I was getting a bug while working on this issue `Error: Cannot find module '019171a2'` while running with `pnpm bootstrap`. Debugging this issue cause lot's of time.是的，我很想在这个问题上工作。我最近没有更新的原因是因为我在运行 `pnpm bootstrap` 时遇到一个 `Error: Cannot find module '019171a2'` 错误。调试此问题会导致大量时间。
> 
[...]

PR #1255, opened 2024-04-27
Create a function to combine resolveFarmPlugin and resolveAsyncPlugins in index.ts

DeepeshKalura:
## Source 
from Issue #1224, "support receive Promise plugin"

## Understand
Combine two function  

## Description of changes:
The `resolveFarmPlugins` function was originally responsible for processing an array of plugins from a configuration object. It would separate these plugins into two categories: Rust plugins and JavaScript plugins. The function would then return an object containing these two arrays.

The `resolveAsyncPlugins` function was responsible for resolving an array of asynchronous plugins. It would return a new array where all asynchronous plugins have been resolved.

In the combined function resolveFarmPluginsAndAsyncPlugins, we first process the plugins as we did in the resolveFarmPlugins function. However, instead of immediately returning the Rust and JavaScript plugins, we pass each array to the resolveAsyncPlugins function. This ensures that all plugins are resolved before they are returned.

The main difference between the original resolveFarmPlugins function and the combined resolveFarmPluginsAndAsyncPlugins function is that the latter ensures all plugins are resolved before they are returned. This is important if any of the plugins are asynchronous and need to be awaited before they can be used.

This combined function provides a more complete solution for processing and resolving plugins, as it handles both the separation of plugin types and the resolution of asynchronous plugins.

### DIFF
--- a/packages/core/src/plugin/index.ts
+++ b/packages/core/src/plugin/index.ts
@@ -1,7 +1,5 @@
-import { isArray, isObject } from '../utils/index.js';
 import { convertPlugin } from './js/index.js';
 import { rustPluginResolver } from './rust/index.js';
-
 import type { JsPlugin } from './type.js';
 import { ResolvedUserConfig, type UserConfig } from '../config/index.js';
 import merge from '../utils/merge.js';
@@ -9,7 +7,8 @@ import merge from '../utils/merge.js';
 export * from './js/index.js';
 export * from './rust/index.js';
 
-export async function resolveFarmPlugins(config: UserConfig) {
+// combing resolveFarmPlugin and resolveAsyncPlugin
+export async function resolveFarmPluginsAndAsyncPlugins(config: UserConfig) {
   const plugins = config.plugins ?? [];
 
   if (!plugins.length) {
@@ -20,7 +19,6 @@ export async function resolveFarmPlugins(config: UserConfig) {
   }
 
   const rustPlugins = [];
-
   const jsPlugins: JsPlugin[] = [];
 
   for (const plugin of plugins) {
@@ -30,15 +28,15 @@ export async function resolveFarmPlugins(config: UserConfig) {
 
     if (
       typeof plugin === 'string' ||
-      (isArray(plugin) && typeof plugin[0] === 'string')
+      (Array.isArray(plugin) && typeof plugin[0] === 'string')
     ) {
       rustPlugins.push(
         await rustPluginResolver(plugin as string, config.root ?? process.cwd())
       );
-    } else if (isObject(plugin)) {
+    } else if (typeof plugin === 'object') {
       convertPlugin(plugin as unknown as JsPlugin);
       jsPlugins.push(plugin as unknown as JsPlugin);
-    } else if (isArray(plugin)) {
+    } else if (Array.isArray(plugin)) {
       for (const pluginNestItem of plugin as JsPlugin[]) {
         convertPlugin(pluginNestItem as JsPlugin);
         jsPlugins.push(pluginNestItem as JsPlugin);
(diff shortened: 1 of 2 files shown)


ErKeLost:
Thanks, but should not update pnpm.lock should use the correct version

ErKeLost:
Because the code of core has been modified, you need to execute `npx changeset` in the `root directory` to update the package of the first version of patch.

packages/core/src/plugin/index.ts
```diff
@@ -1,15 +1,14 @@
...
 import type { JsPlugin } from './type.js';
 import { ResolvedUserConfig, type UserConfig } from '../config/index.js';
 import merge from '../utils/merge.js';
 
 export * from './js/index.js';
 export * from './rust/index.js';
 
-export async function resolveFarmPlugins(config: UserConfig) {
+// combing resolveFarmPlugin and resolveAsyncPlugin
+export async function resolveFarmPluginsAndAsyncPlugins(config: UserConfig) {
```
shulandmimi, line 11: `packages/core/src/config/index.ts:13` references it but does not sync the changes

shulandmimi, line 11: `packages/core/src/config/index.ts:13` references it but does not sync the changes

packages/core/src/plugin/index.ts
```diff
@@ -50,14 +48,16 @@ export async function resolveFarmPlugins(config: UserConfig) {
...
-    rustPlugins,
-    jsPlugins
+    rustPlugins: resolvedRustPlugins,
+    jsPlugins: resolvedJsPlugins
   };
 }
 
-// resolve promise plugins
-export async function resolveAsyncPlugins<T>(arr: T[]): Promise<T[]> {
+async function resolveAsyncPlugins<T>(arr: T[]): Promise<T[]> {
```
shulandmimi, line 60: Same as above, `packages/core/src/config/index.ts:13` references it.

packages/core/src/plugin/index.ts
```diff
@@ -30,15 +28,15 @@ export async function resolveFarmPlugins(config: UserConfig) {
...
     if (
       typeof plugin === 'string' ||
-      (isArray(plugin) && typeof plugin[0] === 'string')
+      (Array.isArray(plugin) && typeof plugin[0] === 'string')
     ) {
       rustPlugins.push(
         await rustPluginResolver(plugin as string, config.root ?? process.cwd())
       );
-    } else if (isObject(plugin)) {
+    } else if (typeof plugin === 'object') {
```
shulandmimi, line 36: typeof is not accurate, for example, a null is received

packages/core/src/plugin/index.ts
```diff
@@ -20,7 +19,6 @@ export async function resolveFarmPlugins(config: UserConfig) {
   }
 
   const rustPlugins = [];
-
   const jsPlugins: JsPlugin[] = [];
 
   for (const plugin of plugins) {
```
shulandmimi, line 24: This change should support async plugin, that is, the plugin may be a promise, so we should judge here whether it is a promise received, wait for it to resolve, and then judge whether it is a rust plugin or a js plugin.

> But the above implementation alone does not support nesting, so it would be better if you could help implement nesting.

shulandmimi (changes requested):

DeepeshKalura:
@ErKeLost Thanks for letting me know, I will revert changes in the `pnpm version` and execute `npx changeset`.  update this PR.

packages/core/src/plugin/index.ts
DeepeshKalura: Could you elaborate the what you mean by references it?

My understanding, you call the function it has synchronized issue.

shulandmimi: You changed its name but did not change it synchronously in `packages/core/src/config/index.ts` where it is used.

- https://github.com/farm-fe/farm/blob/main/packages/core/src/config/index.ts#L7-L14
- https://github.com/farm-fe/farm/blob/main/packages/core/src/config/index.ts#L168-L172

packages/core/src/plugin/index.ts
DeepeshKalura: @shulandmimi you are correct that `typeof` in js can be misleading, especially when dealing with null values, as `typeof` null returns 'object'.  Can you suggest me what to do when error occur.

shulandmimi:
```suggestion
    } else if (isObject(plugin)) {
```

DeepeshKalura:
@shulandmimi Closing this PR and I later will recreate this PR.

PR #1269, opened 2024-04-30
feat: rust plugin usage

### LINKED ISSUE #1213: Normalize usage of rust plugins and js plugins

Configuring rust plugins and js plugins in farm.config.ts uses different methods currently, for example:

```ts
import farmSassPlugin from '@farmfe/js-plugins-sass';

export default defineConfig({
  plugins: [
    // rust plugin
    ['@farmfe/plugin-react', {}], 
    // js plugin
    farmSassPlugin({
      // globals: ['./src/variables.scss']
    })
  ],
});
```
Farm uses `@farmfe/plugin-react` to configure rust plugins and uses `farmSassPlugin({})` to configure js plugins. It's in consistent and it's hard to add types for rust plugin options.

we need to normalize this 2 methods.

```ts
import react from '@farmfe/plugin-react';
import farmSassPlugin from '@farmfe/js-plugins-sass';

export default defineConfig({
  plugins: [
    react({}),
    farmSassPlugin({
      // globals: ['./src/variables.scss']
    })
  ],
});
```
using new `react({})` strategy to configuring rust plugin, and keep the old ways compatible.

callqh:
**Description:**
Make the rust plugin and the js plugin consistent in the way they are used, with type hints.
```ts
import react from '@farmfe/plugin-react';
import farmSassPlugin from '@farmfe/js-plugins-sass';

export default defineConfig({
  plugins: [
    react({}),
    farmSassPlugin({
      // globals: ['./src/variables.scss']
    })
  ],
});
```

**BREAKING CHANGE:**

close #1213


rust-plugins/react/index.js
```diff
@@ -119,4 +119,4 @@ switch (platform) {
     throw new Error(`Unsupported OS: ${platform}, architecture: ${arch}`);
 }
 
-export default binPath;
+export default (options)=>[binPath,options];
```
wre232114, line 122: This would introduce break change when using the new `@farmfe/plugin-react` in old `@farmfe/core`, a new package entry should be added to avoid break change

wre232114 (changes requested):

wre232114 (approved):

PR #1284, opened 2024-05-04
chore(refactor): optimize tree shake implemetation to improve performance

wre232114:
…ance

**Description:**

**BREAKING CHANGE:**

#1276


crates/plugin_tree_shake/src/module.rs
```diff
@@ -221,458 +224,134 @@ impl TreeShakeModule {
...
+  fn all_exports_to_statement_idents(&self) -> Vec<(UsedStatementIdent, StatementId)> {
+    let mut used_idents = vec![];
 
-          for index in removed_idents {
-            item.read_vars.remove(index);
+    for export_info in self.exports() {
+      for sp in &export_info.specifiers {
+        match sp {
+          ExportSpecifierInfo::Default => {
+            used_idents.push((UsedStatementIdent::Default, export_info.stmt_id));
```
shulandmimi, line 240: `export * from` should not contain `default`

[MDN](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export)

> This re-exports all named exports from mod as the named exports of the current module, but the default export of mod is not re-exported.

shulandmimi (changes requested):

shulandmimi (approved):

PR #1380, opened 2024-06-11
fix: doesn't generate the plugin name when copying js plugin template

Weeken:
**Description:**

**BREAKING CHANGE:**


.changeset/ninety-students-call.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-farm-plugin": major
```
wre232114, line 2: It should be `patch` instead of `major`

wre232114 (changes requested):

packages/create-farm-plugin/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "create-farm-plugin",
-  "version": "0.1.7",
+  "version": "0.1.8",
```
wre232114, line 3: should not modify version manually

wre232114 (changes requested):

wre232114 (approved):

PR #1401, opened 2024-06-14
ci: use pnpm cache in workflows

Jayllyz:
**Description:**

Noticed that pnpm was installed with npm, with setup-pnpm before installing Node we can benefit from the cache. I also brought up a few actions that were obsolete.

*First contribution by the way, thanks for creating Farm, it's awesome.*

**BREAKING CHANGE:**


wre232114:
Thanks for the contribution. I'll merge it when all the check passed. Looking forward to more collaboration

fu050409 (approved):

Issue #1262, opened 2024-04-29
feature: support `server.hmr.overlay` control overlay display


ErKeLost:
### What problem does this feature solve?

add `server.hmr.overlay` boolean 

### What does the proposed API look like?

`overlay?: boolean`

fu050409:
I'll try to work with this

PR #1490, opened 2024-07-01
feat(overlay): support disable overlay

fu050409:
**Description:**

Support `server.hmr.overlay` control overlay display

- Add `server.hmr.overlay` option 
- Transfer `overlay` config to fronted end
- Display overlay only when `server.hmr.overlay` is true

Resolved: #1262


ErKeLost:
Thanks I will check the code later

ErKeLost:
Thanks，This PR may need to wait for a while. I am currently reconstructing the error handling and overlay situation. I will check this PR after it is confirmed.

PR #1495, opened 2024-07-01
fix: className instead of class in tauri > react template

jarrisondev:
**Description:**
The Tauri > React template is using class instead of className in an image tag, which produces an error when running npm run build.

```jsx
// Before
<a href="https://farmfe.org/" target="_blank">
  <img src={FarmLogo} class="logo" alt="Farm logo" />
</a>

// After
<a href="https://farmfe.org/" target="_blank">
  <img src={FarmLogo} className="logo" alt="Farm logo" />
</a>
...
```

### DIFF
--- a/.changeset/big-poets-sell.md
+++ b/.changeset/big-poets-sell.md
@@ -1,5 +1,0 @@
----
-"@farmfe/core": patch
----
-
-bump dotenv version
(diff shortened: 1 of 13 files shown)


wre232114:
Thank you! Please add a changeset for `create-farm` package by running `npx changeset`

PR #1500, opened 2024-07-02
fix(template): replace 'class' with 'className' in image tag to resol…

jarrisondev:
**Description:**

The Tauri > React template is using class instead of className in an image tag, which produces an error when running npm run build.

```jsx
// Before
<a href="https://farmfe.org/" target="_blank">
  <img src={FarmLogo} class="logo" alt="Farm logo" />
</a>

// After
<a href="https://farmfe.org/" target="_blank">
  <img src={FarmLogo} className="logo" alt="Farm logo" />
</a>
```

**BREAKING CHANGE:**

### DIFF
--- a/.changeset/nervous-peaches-reply.md
+++ b/.changeset/nervous-peaches-reply.md
@@ -1,5 +1,5 @@
 ---
-"create-farm": minor
+"create-farm": patch
 ---
 
 Corrected the usage of class to className in an image tag within the Tauri > React template to resolve build errors.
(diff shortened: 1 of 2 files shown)


.changeset/nervous-peaches-reply.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-farm": minor
```
wre232114, line 2: bump it as `patch` instead of `minor`

wre232114 (changes requested):

wre232114 (approved):

ErKeLost:
thanks

PR #1511, opened 2024-07-03
fix: resolve absolute path error in the alias options for css file

### LINKED ISSUE #1467: [Bug Report]: resolve absolute path error in the alias options
### Steps to reproduce

```
resolve： {
  alias: {
     '/@': path.resolve('src')
  }
}
```

### Reproduce link

_No response_

resolve absolute alias error

### System Info

```Shell
mac
```

callqh:
**Description:**

**BREAKING CHANGE:**

close #1467


crates/plugin_css/src/dep_analyzer.rs
```diff
@@ -70,7 +70,6 @@ impl Visit for DepAnalyzer {
 pub fn is_source_ignored(source: &str) -> bool {
   source.starts_with("http://")
     || source.starts_with("https://")
-    || source.starts_with('/')
```
wre232114, line 73: we cannot just remove this condition. for url(/logo.png), the resolver may throw 'can not resolve' error, but the image may exist under public dir, in this situation /logo.png should not be handled.

if think we should add more judgement in  resolve hook of plugin css, try alias resolve first but external unresolved /logo.png

wre232114 (changes requested):

crates/plugin_css/src/dep_analyzer.rs
callqh: Yes, I am trying to use the configuration in alias to determine whether further processing is needed.

crates/plugin_css/src/source_replacer.rs
```diff
@@ -59,6 +62,8 @@ impl<'a> VisitMut for SourceReplacer<'a> {
                   } else {
                     self.public_path.trim_end_matches('/')
                   };
+                  println!("需要处理的路径{:?}==={:?}", source, normalized_public_path);
```
wre232114, line 65: remove println

crates/plugin_css/tests/analyze_deps.rs
```diff
@@ -67,6 +77,7 @@ fn analyze_deps() {
     };
 
     css_plugin.analyze_deps(&mut params, &context).unwrap();
+    println!("{:?}", params.deps);
```
wre232114, line 80: remove println

crates/plugin_css/src/dep_analyzer.rs
```diff
@@ -70,7 +76,19 @@ impl Visit for DepAnalyzer {
...
+    || source.starts_with("/")
     || source.starts_with("data:")
     || source.starts_with('#')
 }
+
+pub fn is_start_with_alias(alias: &HashMap<String, String>, source: &str) -> bool {
```
wre232114, line 84: alias may be a regex, refer to plugin resolve, i think a common alias matching util may need to be extracted to toolkit

wre232114 (changes requested):

wre232114 (approved):

Issue #1123, opened 2024-04-04
refactor: `@farmfe/cli`


ErKeLost:
### What problem does this feature solve?

Use more modern tools to optimize farm cli functionality as well as functional programming and performance issues

1. Use [`citty` ](https://github.com/unjs/citty) rewrite cli
2. Split functional structure `watch` , `build`, `start`, `preview`, `clean` command
3. Rewrite cli with more elegant functional programming

### What does the proposed API look like?

detail see [citty](https://github.com/unjs/citty/tree/main/playground)

fu050409:
Working on this issue

PR #1517, opened 2024-07-03
refactor(cli): use `citty` rewrite cli

fu050409:
**Description:**

- Use `citty` rewrite cli
- Refactor all commands to citty
- Defaults to start a dev server

Resolved: #1123


fu050409:
Some checks failed, it seems some mistakes happened in my `build` codes, let me try to fix this

ErKeLost:
Thanks. I will check the code later.

packages/cli/src/index.ts
```diff
@@ -12,183 +12,285 @@ import {
...
-      const defaultOptions = {
-        root,
-        configPath,
-        ...getOptionFromBuildOption(options)
-      };
-
-      const { build } = await resolveCore();
-      handleAsyncOperationErrors(build(defaultOptions), 'error during build');
+  },
+  async run({ args }: { args: FarmCLICommonOptions & FarmCLIServerOptions }) {
```
ErKeLost, line 51: The alias parameter in arg is not configured. Is there a way to configure shared arg in citty? 
If I use -m now it doesn't work

packages/cli/src/index.ts
```diff
@@ -12,183 +12,285 @@ import {
...
-        configPath,
-        mode: options.mode
-      };
-
-      const { start } = await resolveCore();
-      handleAsyncOperationErrors(
-        start(defaultOptions),
-        'Failed to start server'
-      );
+const devCommand = defineCommand({
```
ErKeLost, line 27: The original idea was to cut files with all commands, start command, watch build preview, create new command, create different files in the folder and write them.

packages/cli/src/index.ts
```diff
@@ -12,183 +12,285 @@ import {
...
+    name: 'farm',
+    version
+  },
+  args: {
+    config: {
+      type: 'string',
+      alias: 'c',
+      description: 'use specified config file'
+    },
+    mode: { type: 'string', alias: 'm', description: 'set env mode' },
```
ErKeLost, line 277: shot name it doesnt work

packages/cli/src/index.ts
```diff
@@ -197,8 +299,7 @@ cli.on('command:*', async () => {
 // We only keep the original code environment.
 preventExperimentalWarning();
```
ErKeLost, line 300: Can this method help me delete it? We don’t need this now.

packages/cli/src/index.ts
```diff
@@ -197,8 +299,7 @@ cli.on('command:*', async () => {
...
 
-cli.help();
-
-cli.version(version);
-
-cli.parse();
+// default to start a development server
+if (process.argv.slice(2).length === 0)
```
ErKeLost, line 303: There is a problem with this way of writing. If I follow arg, it will also be used as a parameter. If I use farm -m directly, it will not work.

ErKeLost: I think this way of writing is not good. Doesn’t citty provide this direct operation method?

packages/cli/src/index.ts
fu050409: OK, I'll refactor this later

packages/cli/src/index.ts
fu050409: I'll check this later, but I truely dealed with this issue with this funciton: https://github.com/farm-fe/farm/pull/1517/files#diff-852d90245bbe147114049b7940720369a5dd1c564e5d1ab29b1bf8eae193cef3R192-R207

packages/cli/src/index.ts
fu050409: > I think this way of writing is not good. Doesn’t citty provide this direct operation method

`citty` doesn't provide a default behavior, it means if I run farm-cli with empty args (for example just `pnpm farm`), `citty` will throw an error to tell you that no args provided. But the current version of farm-cli will start a dev server by default.

I tried to add a default handler in `main`, but `citty` will run it twice with subcommand given. This means even you are running `pnpm farm clean`, a dev server will be started soon after cleaned.

packages/cli/src/index.ts
fu050409: Of course, I'll delete it later

fu050409:
@ErKeLost Hi! I created three issues in `citty` (https://github.com/unjs/citty/issues/152, https://github.com/unjs/citty/issues/153 and https://github.com/unjs/citty/issues/154)。 

Do you think we should manually deal with the problem caused by the global parameter first, or do you think we should wait for upstream to implement there features before we deal with this pr?

ErKeLost:
> @ErKeLost Hi! I created three issues in `citty` ( [unjs/citty#152](https://github.com/unjs/citty/issues/152), [unjs/citty#153](https://github.com/unjs/citty/issues/153) and [unjs/citty#154](https://github.com/unjs/citty/issues/154) )。
> 
> Do you think we should manually deal with the problem caused by the global parameter first, or do you think we should wait for upstream to implement these features before we deal with this pr?

I think we can wait for a while. We originally wanted to change city. One is speed and another very important point is functional style and elegance of code. Now it seems that it has deviated from the original intention. We don’t need to do a lot of compatibility to complete it. This cli is unreasonable. We can see if citty is willing or supports these functions.

fu050409:
OK, I had submitted a pull request to solve one of the issues and waiting for their review. I'll continue on this pull request once these issues of `citty` are implemented.

fu050409:
@ErKeLost Hi! I found that citty have not maintained for a long time, so I started a new project named [`archons`](https://github.com/noctisynth/archons) built by napi. This project is based on `clap-rs` to build fast, powerfull and elegant CLI tools. I will next try to migrate our CLI to `archons` instead of `citty` and keep the farm cli adapted to archons!

ErKeLost:
I saw this project which is very interesting. The most important thing about whether we can replace the standard is stability. I can listen to @wre232114  thoughts.

fu050409:
> I saw this project which is very interesting. The most important thing about whether we can replace the standard is stability. I can listen to @wre232114 thoughts.

I have tested on all different platforms with env of node 18 and node 20, it works properly. Of course I'm still not sure that napi will work stably, but if you have the time I'm glad if you can try it out 💯

Issue #1522, opened 2024-07-03
Electron + react template: Failed to construct 'WebSocket': The URL 'ws://localhost:9000__hmr' is invalid. [Bug Report]: 


GustavoGomez092:
### Steps to reproduce

`npm create farm@latest my-react-app`
Then select electron and React template

### What is actually happening?

HMR not working after clean `npm install` and `npm run dev`
with error: 

<p style="color:red;">Uncaught DOMException: Failed to construct 'WebSocket': The URL 'ws://localhost:9000__hmr' is invalid.</p>

<img width="1178" alt="image" src="https://github.com/farm-fe/farm/assets/45272016/d8348fd6-8967-4d81-b50a-6ac3ff76bb5f">

### System Info

```Shell
System:
    OS: macOS 14.5
    CPU: (11) arm64 Apple M3 Pro
    Memory: 552.19 MB / 18.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.14.0 - /usr/local/bin/node
    npm: 10.7.0 - /usr/local/bin/npm
    pnpm: 9.4.0 - ~/Library/pnpm/pnpm
  Browsers:
    Chrome: 126.0.6478.127
    Safari: 17.5
    Safari Technology Preview: 17.4
  npmPackages:
    @farmfe/core: ^1.2.4 => 1.2.6
```

ErKeLost:
i'll try it

PR #1542, opened 2024-07-06
fix(hmr): auto add `/` in socket URL if missing

fu050409:
Resolved: #1522


packages/runtime-plugin-hmr/src/hmr-client.ts
```diff
@@ -13,7 +13,9 @@ const hmrHost =
     : FARM_HMR_HOST || 'localhost';
 
 const socketProtocol = location.protocol === 'https:' ? 'wss' : 'ws';
-const socketHostUrl = `${hmrHost}:${hmrPort}${FARM_HMR_PATH}`;
+const socketHostUrl = FARM_HMR_PATH.startsWith('/')
```
ErKeLost, line 16: under what circumstances will the problem occur/lost?

fu050409: > under what circumstances will the problem occur/lost?

See #1522, socket host url is invalid in electron

ErKeLost: I haven't seen the specific reason why the error was reported, but I think the solution is to find out why / was killed, rather than refixing the problem in the end.

fu050409: > I haven't seen the specific reason why the error was reported, but I think the solution is to find out why `/` was killed, rather than refixing the problem in the end.
> 
> Solve problems at their root rather than patching them up

You are right, I'll try to find where inject these globals.

ErKeLost: If it is a problem with electron, can we solve this problem in the electron plugin?  Because I haven’t looked at the specific reason yet, I can only give you ideological suggestions.

fu050409:
The problem is that when dealing with runtime parameters, we have an exception for posix path handling, and './ 'will be ignored. Also, there is a problem with the section that displays the URL, which displays `http://127.0.0.1:9000\`, I fixed it.

packages/core/src/config/index.ts
```diff
@@ -415,7 +415,9 @@ export async function normalizeUserCompilationConfig(
     const publicPath = resolvedCompilation.output.publicPath;
     const hmrPath = resolvedUserConfig.server.hmr.path;
     const serverOptions = resolvedUserConfig.server;
-    const defineHmrPath = normalizeBasePath(path.join(publicPath, hmrPath));
+    const defineHmrPath = normalizeBasePath(
```
ErKeLost, line 418: There is a problem with using path.join here. We cannot splice / at will.

packages/core/src/config/index.ts
```diff
@@ -415,7 +415,9 @@ export async function normalizeUserCompilationConfig(
     const publicPath = resolvedCompilation.output.publicPath;
```
ErKeLost, line 415: The problem lies here. The public path needs to be formatted. Directly modify this method and wrap it with the getValidPublicPath method.

packages/core/src/config/normalize-config/normalize-output.ts
```diff
@@ -259,7 +260,11 @@ export function getValidPublicPath(publicPath = '/'): string | undefined {
   if (publicPath.startsWith('/')) {
     validPublicPath = publicPath;
   } else if (publicPath.startsWith('.')) {
-    validPublicPath = path.posix.normalize(path.join('/', publicPath));
+    validPublicPath = path.posix.normalize(
```
ErKeLost, line 263: Use normalizeBasePath method here

ErKeLost:
The key to solving the problem lies in the `normalize-output.ts` file. For the processing of `config.output.publicPath`, you need to add normalized `getValidPublicPath`.

fu050409:
@ErKeLost Ok, I've changed it to take into account your comments.

But there's still something I don't realized. `defineHmrPath` is essentially a URL path, and I think it looks like no problem to mount it after `/` here? In fact, the originally solution in `getValidPublicPath` is join `/` with `publicPath`. Such a modification doesn't seem to cause any issues.

ErKeLost:
**https://github.com/farm-fe/farm/pull/1542#issuecomment-2212474087** check this point

fu050409:
> **[#1542 (comment)](https://github.com/farm-fe/farm/pull/1542#issuecomment-2212474087)** check this point, You should try it with the current code. It doesn't work properly.

I tested `pnpm start` and `pnpm dev` in all the examples mannully and nothing seems to go wrong, it also works fine after merging the main branch and cleaning `pnpm bootstrap` to re-bootstarp my workspace. Idk what's going wrong, could you please provide me more info?

ErKeLost:
> > **[#1542 (comment)](https://github.com/farm-fe/farm/pull/1542#issuecomment-2212474087)** check this point, You should try it with the current code. It doesn't work properly.
> 
> I tested `pnpm start` and `pnpm dev` in all the examples mannully and nothing seems to go wrong, it also works fine after merging the main branch and cleaning `pnpm bootstrap` to re-bootstarp my workspace. Idk what's going wrong, could you please provide me more info?

run electron example it doesnt work

fu050409:
@ErKeLost HMR worked properly now, the issue caused by the normalized hmr path at server side is the same as client side.

ErKeLost (approved):

PR #1552, opened 2024-07-08
Update README.md

fyzhu:
**Description:** fix typo

**BREAKING CHANGE:** no

**Related issue (if exists):** no


ErKeLost:
Thanks, but one more thing that needs to be done is the terminal output npx changeset --empty

PR #1571, opened 2024-07-10
fix: Correct Description, CSS, and resolve "Cannot find module" on npm run build

Estereg:
Something very simple


wre232114 (approved):

wre232114:
Thanks!

Issue #1572, opened 2024-07-10
[Bug Report]: Config not completely updated in HMR


fu050409:
### Steps to reproduce

1. Setup a farm project by using `pnpm create farm@latest`
2. Setup environment by using `pnpm i`
3. Start a dev server by using `pnpm dev` or `pnpm start`
4. For example, modify the config file:

   From:

   ```ts
   server: {
     hmr: {
       overlay: true
     }
   },
   ```

   To:

   ```ts
   server: {
     hmr: {
       overlay: false
     }
   },
   ```

5. Modify any files in `src` to approch an error, in broswer, overlay of the error still displays.

### What is actually happening?

see  above

### System Info

```Shell
any
```

PR #1588, opened 2024-07-11
fix(hmr): fix config reload

fu050409:
**Description:**

- Fix config reload

Resolved: #1572

### DIFF
--- a/packages/core/src/config/index.ts
+++ b/packages/core/src/config/index.ts
@@ -101,7 +101,10 @@ async function getDefaultConfig(
     logger
   );
 
-  resolvedUserConfig.server = normalizeDevServerConfig(inlineOptions, mode);
+  resolvedUserConfig.server = normalizeDevServerConfig(
+    inlineOptions.server,
+    mode
+  );
 
   resolvedUserConfig.compilation = await normalizeUserCompilationConfig(
     resolvedUserConfig,
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -427,7 +427,18 @@ export async function createFileWatcher(
 
       await devServer.close();
       __FARM_GLOBAL__.__FARM_RESTART_DEV_SERVER__ = true;
-      await start(resolvedUserConfig as FarmCLIOptions & UserConfig);
+
+      const defaultOptions = {
+        root: resolvedUserConfig.root,
+        compilation: {
+          lazyCompilation: resolvedUserConfig.compilation.lazyCompilation
+        },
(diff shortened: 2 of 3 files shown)


packages/core/src/index.ts
```diff
@@ -427,7 +427,18 @@ export async function createFileWatcher(
...
-      await start(resolvedUserConfig as FarmCLIOptions & UserConfig);
+
+      const defaultOptions = {
+        root: resolvedUserConfig.root,
+        compilation: {
+          lazyCompilation: resolvedUserConfig.compilation.lazyCompilation
+        },
+        server: resolvedUserConfig.server,
+        clearScreen: resolvedUserConfig.clearScreen,
+        configPath: path.dirname(resolvedUserConfig.configFilePath),
```
ErKeLost, line 438: Is there any difference between this and passing resolveConfig directly? I think only the path has been modified. Then we should adjust the configPath issue when parsing the config.

packages/core/src/config/index.ts
```diff
@@ -101,7 +101,10 @@ async function getDefaultConfig(
     logger
   );
 
-  resolvedUserConfig.server = normalizeDevServerConfig(inlineOptions, mode);
+  resolvedUserConfig.server = normalizeDevServerConfig(
```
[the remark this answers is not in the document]
fu050409: Infact, you can see that `normalizeDevServerConfig` function applies 2 args which are `options: UserServerConfig | undefined` and `mode: string`. Here should be a server config but not entire config.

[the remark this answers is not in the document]
ErKeLost: ohh Is there something wrong with this part? got it

packages/core/src/index.ts
ErKeLost: I think maybe these designs are all problematic. @wre232114  take a look ？

fu050409: > I think maybe these designs are all problematic. @wre232114 take a look ？

I have a similar feeling that perhaps we need to rewrite the configuration part somewhat in the future, but I think this solution is the best we can do without a major rewrite to avoid these issues.

fu050409: A pure restart is safe and causes no more problems, and you don't seem to realize that if we remove a configuration instead of overwriting it with a configuration, it will use the old value instead of the default.

fu050409: I just remove this change, maybe we should fix this in #1589

ErKeLost (approved):

PR #1593, opened 2024-07-11
fix(template): fix tauri template

fu050409:
**Description:**

- Use `farm.config.ts` instead of `vite.config.ts` in Tauri template


.changeset/odd-lions-enjoy.md
```diff
@@ -0,0 +1,6 @@
+---
+"@farmfe-examples/vite-adapter-svelet": patch
```
ErKeLost, line 2: This is not necessary examples

.changeset/odd-lions-enjoy.md
```diff
@@ -0,0 +1,6 @@
...
+"@farmfe/core": patch
```
ErKeLost, line 3: The patch version of create-farm should be released.

fu050409: Thanks it's my mistake, I'll fix this now

ErKeLost (approved):

Issue #1594, opened 2024-07-11
[Bug Report]: Unrecognized key(s) in object: 'protocol' at "server.hmr".


fu050409:
### Steps to reproduce

1. Setup a farm project
2. Edit `farm.config.ts`:

   ```ts
   export default defineConfig({
     vitePlugins: [vue()],
     server: {
       port: 1420,
       strictPort: true,
       host: mobile ? "0.0.0.0" : false,
       hmr: {
         protocol: "ws", // `protocol` was defined in `UserHmrConfig` but got an error
         host: "127.0.0.1",
         port: 1421,
       },
     },
     compilation: {
       watch: {
         ignored: ["**/src-tauri/**"],
       },
     },
   });
   ```

### What is actually happening?

```
> grassator@0.0.0 dev D:\Projects\Grassator
> farm start

[ Farm ] Using config file at D:\Projects\Grassator\farm.config.ts
[ Farm ] Error: Failed to start the server
Caused by: Error: Failed to load farm config file: Validation error: Unrecognized key(s) in object: 'protocol' at "server.hmr". 
 Please check your configuration file or command line configuration.. 
 Potential solutions: 
1. Try set `FARM_CONFIG_FORMAT=cjs`(default to esm)
2. Try set `FARM_CONFIG_FULL_BUNDLE=1` 
 Error: Validation error: Unrecognized key(s) in object: 'protocol' at "server.hmr". 
 Please check your configuration file or command line configuration.
    at parseUserConfig (file:///D:/Projects/Grassator/node_modules/.pnpm/@farmfe+core@1.2.8/node_modules/@farmfe/core/dist/config/schema.js:365:15)
    at loadConfigFile (file:///D:/Projects/Grassator/node_modules/.pnpm/@farmfe+core@1.2.8/node_modules/@farmfe/core/dist/config/index.js:583:35)
    at async resolveConfig (file:///D:/Projects/Grassator/node_modules/.pnpm/@farmfe+core@1.2.8/node_modules/@farmfe/core/dist/config/index.js:62:34)
    at async start (file:///D:/Projects/Grassator/node_modules/.pnpm/@farmfe+core@1.2.8/node_modules/@farmfe/core/dist/index.js:29:36)
    at async handleAsyncOperationErrors (file:///D:/Projects/Grassator/node_modules/.pnpm/@farmfe+cli@1.0.2/node_modules/@farmfe/cli/dist/utils.js:111:9)
[...]

PR #1596, opened 2024-07-11
fix(config): add protocol in config schema

fu050409:
**Description:**

- Add protocol option
- `nonempty()` deprecated and use `min(1)` instead

Resolved: #1594


ErKeLost:
We don't seem to support wss yet. There should be a problem with the code logic.

fu050409:
> We don't seem to support wss yet. There should be a problem with the code logic.

This PR solves the problem that `protocol` option is not declared in the parser. In our type annotation, the `protocol` parameter exists, but is not recognized by the parser, which will throw an error when parsing the configuration file (as long as it is set, regardless of the value). In #1594, I set `protocol` to `ws` and an error occured.

ErKeLost:
But we have not implemented this function. Currently, our server has many functional defects. This is why I want to reconstruct node. This option currently has no effect.

ErKeLost (approved):

Issue #1597, opened 2024-07-11
[Bug Report]: `farm build` nerver return if `compilation.watch` configured


fu050409:
### Steps to reproduce

1. Modify `farm.config.ts`:

   ```ts
   vitePlugins: [vue()],
   compilation: {
       watch: {
           ignored: ["**/src-tauri/**"],
       },
   },
   ```

2. Run `pnpm build`:

   ```bash
   [ Farm ] Using config file at D:\Projects\Grassator\farm.config.ts
   [ Farm ] Build completed in 299ms ⚡️FULL EXTREME! Resources emitted to ./dist.
   [ Farm ] Public directory resources copied successfully.
   ```

### What is actually happening?

`farm build` process is blocked and will never return

### System Info

```Shell
any
```

fu050409:
See #1598 for detail

PR #1598, opened 2024-07-11
fix(build): prevent create file watcher in production mode

fu050409:
**Description:**

- Prevent create file watcher in `production` mode

Resolved: #1597


packages/core/src/index.ts
```diff
@@ -306,7 +306,11 @@ export async function createBundleHandler(
     logger
   );
 
-  if (resolvedUserConfig.compilation?.watch || watchMode) {
```
ErKeLost, line 309: We allow users to configure the development mode to production

fu050409: OK, do you think just close it in `build` after watcher created or do you think to convey a new var to `createBundleHandler` to fix #1597?

fu050409: I found it `watchMode` will only be true in `watch` command, It seems to be ok to just use `watchMode` to judge whether create a watcher?

shulandmimi:
#1597 watch is only enabled, but [the options are not used](https://github.com/farm-fe/farm/blob/main/packages/core/src/watcher/index.ts#L107)

Currently, the watch configuration is not used anywhere. if it is not to supplement the watch configuration, then the changes here are meaningless.

fu050409:
> #1597 watch is only enabled, but [the options are not used](https://github.com/farm-fe/farm/blob/main/packages/core/src/watcher/index.ts#L107)
> 
> Currently, the watch configuration is not used anywhere. if it is not to supplement the watch configuration, then the changes here are meaningless.

This pull request I have submitted is to resolve the fact that the farm build process is blocked during the watch configuration exists. In the integrated environment, configuring the watch option causes the process to fail to continue. You can actually see that the functions wrapped in the code are not reused elsewhere, i.e. only the `watch` and `build` command are used. So it looks like we don't have to consider whether the watch configuration exists or not, we just need to determine the current instruction.

shulandmimi:
> This pull request I have submitted is to resolve the fact that the `farm build` process is blocked during the watch configuration exists. 

#1597 uses an unfinished/unimplemented feature, so it cannot be considered a bug, and it is not necessary to fix it.

Based on the features you need, you should try to use `server.hmr.watchOptions` or implement `compilation.watch`

fu050409:
> > This pull request I have submitted is to resolve the fact that the `farm build` process is blocked during the watch configuration exists.
> 
> #1597 uses an unfinished/unimplemented feature, so it cannot be considered a bug, and it is not necessary to fix it.
> 
> Based on the features you need, you should try to use `server.hmr.watchOptions` or implement `compilation.watch`

Thanks for the description, this looks like a mistake on my part

shulandmimi:
Anyway, thanks for your PR. this is a legacy issue. i think it can be optimized and should be explained before it to avoid misleading others. This is our mistake.

PR #1606, opened 2024-07-15
fix(wip): Support #[cache_item] in external rust plugins

wre232114:
**Description:**

**BREAKING CHANGE:**


wre232114:
Deprecated, we should not rely on rkyv

PR #1629, opened 2024-07-17
chore: improve partial bundle group named

### LINKED ISSUE #1602: [Feature Request]: Optimize partial bundling

Currently we can not get any bundle info from the output file, all output files are like `index_[hash].[hash].js`. We need to provide some info when configuring `groups` or bundle dependencies. E.g:
* `index_react_react_dom.[hash].js`
* `index_mybundle.[hash].js` when configuring `groups: { name: 'mybundle' }`

1. Support `[entryName]` placeholder for `output.filename`, `[entryName]` and `contentHash`
2. Add `name` to output filename when `partialBundling.groups` is configured
3. Support `enforce` in `groups` to prevent the configured group merged with other groups

shulandmimi:
**Description:**

**BREAKING CHANGE:**

close #1602

### DIFF
--- a/crates/core/src/config/partial_bundling.rs
+++ b/crates/core/src/config/partial_bundling.rs
@@ -59,6 +59,8 @@ pub struct PartialBundlingGroupConfig {
   pub group_type: PartialBundlingGroupConfigGroupType,
   /// all, initial, async
   pub resource_type: PartialBundlingGroupConfigResourceType,
+  /// disable group merge
+  pub enforce: bool,
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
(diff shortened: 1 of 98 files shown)


crates/core/src/config/partial_bundling.rs
```diff
@@ -59,6 +59,8 @@ pub struct PartialBundlingGroupConfig {
   pub group_type: PartialBundlingGroupConfigGroupType,
   /// all, initial, async
   pub resource_type: PartialBundlingGroupConfigResourceType,
+  /// disable group merge
+  pub enforce: bool,
```
wre232114, line 63: This would introduce rust plugin break change, use config.custom instead

wre232114 (approved):

Issue #1497, opened 2024-07-01
[Feature Request]: Support `viteConfig['configFile']` to allow use of `vanilla-extract`'s vite plugin


spikecodes:
### What problem does this feature solve?

Cannot use `@vanilla-extract/vite-plugin` because if you do, you get this error message:

```
[ Farm ] Failed to start the server: 
 Error: Vite plugin 'vanilla-extract' is not compatible with Farm for now. Because it uses viteConfig['configFile'] which is not supported by Farm. Current supported keys are:
root,base,publicDir,mode,define,command,resolve,plugins,server,isProduction,css,build,logger,cacheDir,envDir,assetsInclude,legacy,optimizeDeps,ssr,logLevel,experimental,test,clearScreen,customLogger
```

### What does the proposed API look like?

Internally implement `viteConfig['configFile']`.

PR #1636, opened 2024-07-18
fix: normalize root path for vite plugins compatibility

fu050409:
**Description:**

In #1497, vite plugin named `@vanilla-extract/vite-plugin` uses absolute path to judge if a file path is startswith root path, but file path is normalized but root path is not.

Resolved: #1497


ErKeLost (approved):

wre232114:
We can add a example for this plugin to make sure it works

wre232114 (approved):

Issue #1589, opened 2024-07-11
Farm Roadmap  v2.0   🎉


ErKeLost:
## This refactoring will ensure that a large number of api and property break changes will not be designed

1. Optimize existing @farmfe/core config logic design

2. Add debug debugging to the node layer

3. Replace koa with connect

4. Try to adapt some hmr, jsPlugin Hook, etc. to farm native servers, rather than putting pressure on farm's vite compatibility layer, which will cause excessive pressure on subsequent maintenance

5. The middleware design needs to be re-considered to adapt to the existing koa middleware and user-written middleware or the reimplementation of other native middleware that needs to be replaced.

Currently, we are still thinking about design and finally achieve the goal of farm development rust ecosystem, and js ecosystem reuses vite ecosystem.

Now considering the initial stage, we are still in the process of design thinking, and may mainly consider it in the mojor version

## Reasons for refactoring this node layer

Only by making the underlying infrastructure good enough can it better serve the stability of the upper framework.

This major and important update is mainly to accompany the sustainable development of the vite ecosystem, so that farm can follow the footsteps of vite's js ecosystem for a long time to come.

## Progress till MVP ![](https://geps.dev/progress/60)

# New Docs

- [x] [new Docs](https://farm-neightly.netlify.app/)

# Refctor functions

### `vite adapter`

Refactoring all vite proxy logic mainly targets errors caused by invalid configurations such as `esbuild` ,`rollup` options
Elegant optimized performance
Compatible with more vite plugins

- [x] #1788
- [ ] #1846

## `@farmfe/cli`

- [x]  refactor cli (Without affecting the operation of coredao, all functions of this package have been reconstructed.)
   - [x] Optimize cli logic to make cli logic clearer and more readable
   - [x] refactor start command
   - [x] refactor build command
   - [x] refactor watch command
   - [x] refactor preview command
   - [x] optimize clean command
   - [x] All api method attribute configuration types are rewritten
   - [x] remove watch command save dev and build

### `DX`
- [x] refactor handle error
- [x] refactor overlay page

### `Config`
- [x] Correlate all error return messages
- [ ] provide cli shortcuts
- [x] add `server.watch` options remove `hmr.watch`
- [x] add `cacheDir` options
- [x] clearScreen logic
- [x]  rewrite all types
- [x]  refactor logger 
   - [x] Optimize the usage logic of logger and the logic passed to different methods
- [x]  add debug Set debug mode
- [ ] All api method attribute configuration types are rewritten
[...]

PR #1648, opened 2024-07-19
refactor: node

ErKeLost:
[see #1589](https://github.com/farm-fe/farm/issues/1589)


ErKeLost:
prepare test preview version

PR #1666, opened 2024-07-25
refactor(watcher): optimize file watch and support watch `node_modules`

### LINKED ISSUE #949: Optimize file watch when adding modules in HMR

Currently If a HMR adds a lot of modules, all these modules will be added to the watch list.

We should only watch the common root of these modules

1. record watched paths
2. find common root of the watched paths instead of watching all added modules paths

### LINKED ISSUE #1212: Support watch files under node_modules

Sometimes we need to modify files under node modules and enabled hmr for it

support configuring watch node modules:
image

fu050409:
Resolved #1212, Resolved #949

### DIFF
--- a/.changeset/little-plums-type.md
+++ b/.changeset/little-plums-type.md
@@ -1,0 +1,5 @@
+---
+"@farmfe/core": patch
+---
+
+Support watch files in `node_modules/`
--- a/.changeset/neat-bees-trade.md
+++ b/.changeset/neat-bees-trade.md
@@ -1,0 +1,5 @@
+---
+"@farmfe/core": patch
+---
+
+Optimize file watch logic when adding file in HMR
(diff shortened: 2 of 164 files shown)


ErKeLost:
The watcher may have to wait until the server and hmr functions are completed before it can be started.

The watcher should be configured according to vite. Maybe we can reconstruct the watch node mode first ？

fu050409:
> The watcher may have to wait until the server and hmr functions are completed before it can be started.
> 
> The watcher should be configured according to vite. Maybe we can reconstruct the watch node mode first ？

Yes, I am going to do this. And I am now trying to resolve #949 first and then I will refactor the watcher login to make it works like vite.

fu050409:
In fact, as you can see in #1598, we need to do much more than that in terms of refactoring the watcher. There are still some unimplemented features in the watcher.

fu050409:
Changes all resolved in #1735, close this pull request.

PR #1667, opened 2024-07-25
chore: Improve binary performance and codspeed ci

ErKeLost:
**Description:**

**BREAKING CHANGE:**


fu050409 (approved):

fu050409:
I see that you've changed the optimisation level from `s` to `3`, which does drastically improve the speed of Rust programs (I tested this in another project, with `opt-level=3` can improve the performance as the number of between `20%~60%`, in few conditions will regress the performance), but at the cost of doubling the compile time and even more. I think maybe you should probably create a separate command for continuous integration to compile and test ?

ErKeLost:
I changed the configuration of test and used release and it crashed directly.

PR #1700, opened 2024-08-07
refactor(create-farm): rewrite `create-farm`

fu050409:
**Description:**

**BREAKING CHANGE:**


.changeset/.markdownlint.json
ErKeLost, line 1: What is this used for? I don’t see the ci code for markdown.

ErKeLost (approved):

.changeset/.markdownlint.json
fu050409: This is a config for markdown lint, this will let vscode extension ignore `MD041` standard for the changesets files. Of course this file take no exacly effect on the code and the ci pipeline.

For `MD041`, see https://github.com/DavidAnson/markdownlint/blob/v0.34.0/doc/md041.md.

However, if you do think the file is unnecessary, I will delete it.

ErKeLost: We shouldn't need this for the time being, We currently do not have a vscode plugin set for the current project，think about it later

PR #1735, opened 2024-08-23
refactor(core): node watcher

fu050409:
**Description:**

**BREAKING CHANGE:**


ErKeLost:
The new wacher should work with newServer instead of the existing server

fu050409:
I see. Before it was because I saw that the code in `core`'s `index.ts` wasn't completely implemented (especially `start2` confused me a bit), then I'll rewrite that part of the code next.

PR #1750, opened 2024-08-29
fix: arcgis lazy compile fail

wre232114:
**Description:**

**BREAKING CHANGE:**


crates/compiler/src/update/mod.rs
```diff
@@ -524,6 +569,14 @@ impl Compiler {
       &mut module_group_graph,
     );
 
+    // record diff and patch result
+    self.set_hmr_diff_stats(PrintedDiffAndPatchContext {
+      removed_modules: removed_modules.iter().map(|m| m.0.clone()).collect(),
+      affected_module_groups: affected_module_groups.clone(),
+      diff_result: diff_result.clone(),
+      start_points: start_points.clone(),
+    });
```
shulandmimi, line 578: Is it possible to clone it only after confirming the use?

js-plugins/visualizer/src/node/dataSource.ts
```diff
@@ -21,9 +21,7 @@ export function createDateSourceMiddleware(compiler: Compiler) {
         res.end(JSON.stringify(result));
       };
 
-      if (pathname === '/__record/modules') {
```
shulandmimi, line 24: Is this an expected change? I see the `/__record/modules` interface is still used by some modules.

shulandmimi (approved):

Issue #1769, opened 2024-09-13
[Bug Report]: `yarn create farm` failed


fu050409:
### Steps to reproduce

```console
~\Desktop
❯ yarn create farm myapp --template vue2
➤ YN0000: · Yarn 4.3.1
➤ YN0000: ┌ Resolution step
➤ YN0035: │ create-farm-freebsd-x64@npm:0.0.14: Package not found
➤ YN0035: │   Response Code: 404 (Not Found)
➤ YN0035: │   Request Method: GET
➤ YN0035: │   Request URL: https://registry.yarnpkg.com/create-farm-freebsd-x64
➤ YN0000: └ Completed in 1s 555ms
➤ YN0000: · Failed with errors in 1s 578ms
```

### What is actually happening?

See above. That's odd, I'm executing this command on Windows and not FreeBSD

### System Info

```Shell
System:
    OS: Windows 11 10.0.22631
    CPU: (20) x64 13th Gen Intel(R) Core(TM) i9-13900H
    Memory: 3.70 GB / 15.64 GB
  Binaries:
    Node: 21.6.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 4.3.1 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 10.8.2 - C:\Program Files\nodejs\npm.CMD
    pnpm: 9.10.0 - ~\AppData\Local\pnpm\pnpm.CMD
    bun: 1.1.27 - ~\.bun\bin\bun.EXE
  Browsers:
    Edge: Chromium (127.0.2651.74)
    Internet Explorer: 11.0.22621.3527
```

fu050409:
This also failed on Kali Linux:

![image](https://github.com/user-attachments/assets/112736b7-67be-4b77-a060-28d12cc33a01)

PR #1770, opened 2024-09-13
ci(create-farm): add support for freebsd

fu050409:
Yarn auto resolve dependencies even they won't be installed so `yarn create farm` will fail in any platform since `freebsd-x64` not published but included in `triples.addtional`.

Now we compile `freebsd-x64` only for `create-farm`.

Resolved: #1769


ErKeLost:
@wre232114

wre232114 (approved):

wre232114:
ci build failed

fu050409:
> ci build failed

It caused by an unexpected error by GitHub, it works now after rerun the actions.

PR #1868, opened 2024-10-23
chore(nightly): fix targetEnv schema

fu050409:
TypeScript static type check unable to auto check dyn types.

Declare as `z.enum(Object.values(TARGET_ENV) as [string, ...string[]]).optional(),` will be seen as `z.ZodOptional<z.ZodEnum<[string, ...string[]]>>`.

Use `(typeof TARGET_ENV)[keyof typeof TARGET_ENV]` will fix this issue.


packages/core/src/config/schema.ts
```diff
@@ -50,9 +51,7 @@ const outputSchema = z
     path: z.string().optional(),
     publicPath: z.string().optional(),
     assetsFilename: z.string().optional(),
-    targetEnv: z
-      .enum(Object.values(TARGET_ENV) as [string, ...string[]])
-      .optional(),
+    targetEnv: z.custom<(typeof TARGET_ENV)[keyof typeof TARGET_ENV]>(),
```
ErKeLost, line 54: z.custom 是不是需要返回一个函数

ErKeLost: https://zod.dev/?id=custom-schemas

fu050409: 用法就是`z.custom<T>()`，不过这里忘了加`optional`了我改一下

fu050409: > https://zod.dev/?id=custom-schemas

哦哦是需要一个验证函数是吧

ErKeLost: 对 我看他说要最好加上这个函数

PR #1887, opened 2024-10-27
feat(plugins): add primitives to support sub-compiler usage

ssttevee:
**Description:**

I'm building framework around farm that requires "sub-build"s. The main problem blocking me from achieving this is the inability to access the plugins from the current compiler context/plugin driver.

With these changes, I am able to create a sub-compiler to create a sub-build like this:

```
fn load(
  &self,
  param: &PluginLoadHookParam,
  context: &Arc<CompilationContext>,
  hook_context: &PluginHookContext,
) -> Result<Option<PluginLoadHookResult>> {
  let mut config = context.config.as_ref().clone();
  config.progress = false;
  config.input = HashMap::new();
  config.input.insert(String::from("client"), "anothermodule.ts");

  let compiler =
    Compiler::new_without_internal_plugins(config, context.plugin_driver.plugins.clone())?;

  // copy module graph to new compiler
  context
    .module_graph
    .read()
    .copy_to(&mut *compiler.context().module_graph.write(), true)?;

  compiler.compile()?;

  let resource_map = compiler.context().resource_map.lock();
  let js = String::from_utf8(resource_map.get("client.js").unwrap().bytes.clone()).unwrap();

  // ...
}
```

Note that there are a few issues with this approach:
- a new thread pool is created for each sub-compiler,
- modules may be cloned twice per sub-compiler, once when creating each sub-compiler and once when updating the main module graph with newly cached modules, and
- my plugin binary becomes more bloated than necessary because the entire compiler crate including all other default plugins and dependencies are compiled and bundled with my plugin logic.

**BREAKING CHANGE:**

None.

### DIFF
--- a/crates/compiler/src/lib.rs
+++ b/crates/compiler/src/lib.rs
@@ -79,6 +79,13 @@ impl Compiler {
 
     plugins.append(&mut plugin_adapters);
 
+    Self::new_without_internal_plugins(config, plugins)
+  }
+
+  pub fn new_without_internal_plugins(
+    config: Config,
+    mut plugins: Vec<Arc<dyn Plugin>>,
+  ) -> Result<Self> {
     // sort plugins by priority to make larger priority plugin run first
     plugins.sort_by_key(|b| std::cmp::Reverse(b.priority()));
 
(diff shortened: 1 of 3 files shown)


wre232114 (approved):

wre232114:
Thanks for the PR. We'll support compiling dynamically added inputs in Farm v2.0, it can solve the issue you mentioned, sharing thread pool and module graph.

wre232114:
There are CI errors @ssttevee

fu050409 (approved):

PR #1891, opened 2024-10-28
chore(v2): Integrate middleware and htmlfallback middlewear usage

ErKeLost:
Integrating farm dev server has `all middleware` functions and the impact of `appType` on configuring `spa`, `mpa` and `custom` fields functions

The following is a brief explanation of the core points of this pr, only about the appType function

image

The functions of each part of the middleware in the following order are as follows

- corsMiddleware: Allowing the server to control which cross-domain requests can be accepted by the browser
- proxyMiddleware: Mainly used to process proxy requests in the development server
- publicPathMiddleware: Routing control to be used to handle static resource and URL redirects
- adapterViteMiddleware: A middleware used to process resource files in Vite development mode. Its main purpose is to ensure that certain resources (such as images or fonts) are compiled and provided correctly in development mode
- lazyComplationMiddleware: Handle routing lazy loading middleware
- resourceMiddleware: Handle all resouce to change browser url
- publicMiddleware: Is a middleware used to process static resources, mainly used to provide static files in public directories in the development server
- notFoundMiddleware: return 404
- htmlFallbackMiddleware: Returns different html benchmark resources for different source and apptype return values

The following functional areas are divided into two points
1. All middleware is selected by default
2. If apptype is set to custom, `htmlFallbackMiddleware` and `notFoundMiddleware` will not be used

if the appType is set to `spa`, but the user passes in multiple inputs at this time
```ts
import { defineConfig } from "@farmfe/core";
import path from "path";

export default defineConfig({
  compilation: {
    input: {
      index: path.resolve(__dirname, "index.html"),
      base: path.resolve(__dirname, 'base.html'),
      about: path.resolve(__dirname, 'about.html'),
    },
  },
  server: {
    appType: "mpa",
  },
});
```

Then the following paths are valid as

### if the appType is set to `spa`,

effective：
`localhost:3000`
`localhost:3000/`
`localhost:3000/123456`
`localhost:3000/index.html`
`localhost:3000/base.html`
`localhost:3000/about.html`

### if the appType is set to `mpa`

this `localhost:3000/123456` will be not effective return undefined

### if the appType is set to `custom`

```ts
    if (appType === 'spa' || appType === 'mpa') {
      this.middlewares.use(htmlFallbackMiddleware(this));

      this.middlewares.use(notFoundMiddleware());
    }
```

Then we will not control the situation where middlewareMode is currently only set to true in the user's functional scenario for returning html

Subsequent pr will analyze all scenarios and functions of each middleware in detail

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,7 +4,7 @@
   "biome.enabled": true,
   "editor.defaultFormatter": "biomejs.biome",
   "[typescript]": {
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
+    "editor.defaultFormatter": "vscode.typescript-language-features"
   },
   "[javascript]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
--- a/.changeset/loud-dogs-double.md
+++ b/.changeset/loud-dogs-double.md
@@ -1,0 +1,5 @@
+---
+"@farmfe/core": major
+---
+
+collect htmlFallbackMiddleware
--- a/examples/refactor-react/base.html
+++ b/examples/refactor-react/base.html
@@ -9,6 +9,6 @@
   <title>Farm + React + TS</title>
 </head>
 <body>
-  <div id="root">dfgdfgfdgdfgdfgdfgdfg</div>
+  <div id="root">asdsadasdasdasd</div>
 </body>
 </html>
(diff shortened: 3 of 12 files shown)


.vscode/settings.json
```diff
@@ -4,7 +4,7 @@
   "biome.enabled": true,
   "editor.defaultFormatter": "biomejs.biome",
   "[typescript]": {
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
+    "editor.defaultFormatter": "vscode.typescript-language-features"
```
wre232114, line 7: We should use biome to format code

wre232114 (approved):

PR #1941, opened 2024-11-16
refactor(v2): use faster hasher to optimize

fu050409:
**Description:**

**BREAKING CHANGE:**


wre232114:
You can rename `FxHashMap` to `HashMap` and reexport them at the place. So all usage will not be changed, also it's easy to update the hash implementation in the future

wre232114:
yeah, it's fine

wre232114 (commented):
can we add a new() method for the faster hashmap, then we don't need to replace all the new() to default()

fu050409:
> can we add a new() method for the faster hashmap, then we don't need to replace all the new() to default()

I'm afraid not. The method `new()` is provided by the Rust standard library. And in soruce code, it is:

```
#[inline]
#[must_use]
#[stable(feature = "rust1", since = "1.0.0")]
pub fn new() -> HashMap<K, V, RandomState> {
    Default::default()
}
```

As we can see, this method is for `HashMap` with `RandomState` so it's not implemented for `FxHashBuilder`.

Otherwise, we cannout define inherent `impl` for foreign type.

wre232114 (approved):

Issue #1859, opened 2024-10-22
[Bug Report]: `v2` config command logger should use short path name in windows


ErKeLost:
### Steps to reproduce

version 

    "@farmfe/cli": "2.0.0-nightly-20241022101407",
    "@farmfe/core": "2.0.0-nightly-20241022041556",
    "@farmfe/plugin-react": "2.0.0-nightly-20241022124925",

use example in windows command line should be use shortname

![image](https://github.com/user-attachments/assets/7a95f82a-861b-47da-92f2-4dd991be97e5)

### What is actually happening?

absolute path now

✅ farm.config.ts
❌ /xx/farm/examples/react/farm.config.ts

### System Info

```Shell
windows
```

remloyal:
Execute `Farm Clean`
report errors:
```
[ Farm ] Error: Error cleaning cache in undefined: The "path" argument must be of type string or an instance of Buffer or URL. Received undefined
```
version:

`@farmfe/core: 2.0.0-nightly-20241023020505 `

ErKeLost:
@remloyal   This should be another bug that should be related to parsing cache config  https://github.com/farm-fe/farm/issues/1860

K0nnyaku:
Hi there, I am a new comer, can I try to solve this issue?

fu050409:
Sure, feel free to go ahead!

PR #1944, opened 2024-11-17
fix: short name display on windows

K0nnyaku:
The original code uses `path.posix.relative()` that specifics the platform to linux, use `path.relative` to automaically determine how to extract the relative path.

Resolved: #1859


fu050409:
Thanks for your contribution!

fu050409 (approved):

PR #1991, opened 2024-12-10
fix: some bugs

ErKeLost:
**Description:**

**BREAKING CHANGE:**


packages/core/src/server/index.ts
```diff
@@ -128,6 +128,7 @@ export class Server extends httpServer {
    */
   constructor(readonly inlineConfig: FarmCliOptions & UserConfig) {
     super();
+    this.logger = new Logger();
```
fu050409, line 131: Honestly, I'm not quite sure what this line of code does, it looks like we're correctly creating a logger when we call `createServer` anyway, is it really necessary here?

PR #1997, opened 2024-12-16
feat: support js process module hook

shulandmimi:
**Description:**

**BREAKING CHANGE:**

### DIFF
--- a/.changeset/twelve-radios-hear.md
+++ b/.changeset/twelve-radios-hear.md
@@ -1,5 +1,5 @@
 ---
-"@farmfe/core": patch
+"@farmfe/core": minor
 ---
 
 js plugin supports process module hook
--- a/.changeset/twelve-radios-hear.md
+++ b/.changeset/twelve-radios-hear.md
@@ -1,0 +1,5 @@
+---
+"@farmfe/core": patch
+---
+
+js plugin supports process module hook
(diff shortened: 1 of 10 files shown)


.changeset/twelve-radios-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+"@farmfe/core": patch
```
wre232114, line 2: Make a minor change instead of patch

crates/node/src/plugin_adapters/js_plugin_adapter/hooks/process_module.rs
```diff
@@ -0,0 +1,184 @@
...
+            content: result,
+          },
+          ctx,
+          None,
+        )?
+      else {
+        return Ok(None);
+      };
+
+      convert_code_to_metadata(param, result.content)?;
```
wre232114, line 177: Can we only parse and codegen ast once? we can call all process_module hook of js plugins in serial , only codegen  the ast at start and parse content back when all plugins finished

wre232114 (changes requested):

crates/node/src/plugin_adapters/js_plugin_adapter/hooks/process_module.rs
shulandmimi: Requires a destructive update of the plugin driver, not changed in 1.x

wre232114 (approved):

PR #2051, opened 2025-01-09
feat(compiler): use rust to write resources in parallel

fu050409:
**Description:**

**BREAKING CHANGE:**


ErKeLost:
`write_resource_to_disk` should consider putting the logic of generating the dist directory and the logic of the public directory into rust. I will consider this in the future. I will first incorporate this pr into v2.

ErKeLost:
It seems that it would be better if we were all doing concurrency.

ErKeLost (approved):

PR #2081, opened 2025-02-08
feat: write generate resource and some utils

ErKeLost:
#1793

1. add output file print log
image

3. add writeBundle 
4. prepare out dir logic move to rust side 
    - create out_dir dir
    - copy public dir
    - filter resource.emitter === true
5. add rust side logger
6. add some toolkit utils


crates/compiler/src/write/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+      // TODO: add Farm rust logger
+      println!(
+        "\n(!) Warning: The public directory feature may not work correctly. \
+               outDir {} and publicDir {} are not separate folders.\n",
+        output_dir_path.display(),
+        public_dir_path.display()
+      );
+    }
+
+    self.copy_dir(public_dir_path, output_dir_path);
```
fu050409, line 74: wdyt to add a plugin driver hook to notify plugins about public resource copy? or just leave the plugins to resolve completed files itself?somtimes some elder projects may need users add some files to public dir so we cannot optimize these files for now

package.json
```diff
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "preinstall": "npx only-allow pnpm",
-    "bootstrap": "node scripts/bootstrap.mjs",
+    "bootstrap": "node scripts/bootstrap.mjs -v",
```
fu050409, line 11: i think maybe it's better to use `pnpm bootstrap -v` instead?

ErKeLost: The reason why I suggest using verbose is that in some special cases, such as the first run of farm, Rust build can be extremely slow. Many developers cannot see the real Rust compilation progress and come to ask me if it is stuck. This is also why I use inherit to prompt developers.

crates/compiler/src/write/mod.rs
ErKeLost: Copying publicDir should be our internal behavior of the framework. If the user needs to perform other special operations, they can do so in the `finish` hook or the upcoming `writeBundle` hook.

ErKeLost: After the entire write is completed, we can provide users with a writeBundle hook to manipulate the output file or other files.

package.json
ErKeLost: can add another command, but I still feel that verbose is the user's first intuition.

ErKeLost:
image

Outputting the output file size in the Rust hook causes the program to wait for all threads to finish executing before proceeding, even though multiple threads have been started. This leads to longer overall compilation times. In JavaScript, we could make this run asynchronously, but I wonder if there are any better solutions; otherwise, we'll have to implement it in JavaScript after all.

Wait for all threads to finish execution here.
```ts
    const startTime = performance.now();
    await compiler.compile();
    const elapsedTime = Math.floor(performance.now() - startTime);
    const persistentCacheText = persistentCache
      ? bold(PersistentCacheBrand)
      : '';
```

packages/core/src/config/index.ts
```diff
@@ -165,6 +165,17 @@ export async function resolveConfig(
     sortFarmJsPlugins
   );
 
+  if (config.compilation.mode === 'production' && !config.watch) {
```
wre232114, line 168: add a config option like `output.fileSize` and move this if statement to rust compiler similar to minify or tree shake.

wre232114 (approved):

wre232114 (approved):

PR #2082, opened 2025-02-09
feat: cache module metadata & cachable deserialize

shulandmimi:
**Description:**

module metadata cache（Replace write_plugin_cache, plugin_cache_loaded）

```rs
// write cache
context.cache_manager.module_cache.write_metadata(key, name, value)

// read cache
context.cache_manager.module_cache.read_metadata_ref(key, name)

// cache container for primitive type
CacheableContainer::new(v)
// e.g:
CacheableContainer::new(1)
CacheableContainer::new(String::from("hello world"))
// ...
```

**BREAKING CHANGE:**


crates/plugin_css/src/lib.rs
```diff
@@ -192,26 +192,42 @@ impl Plugin for FarmPluginCss {
...
+    context: &Arc<CompilationContext>,
   ) -> farmfe_core::error::Result<Option<()>> {
-    let cache = deserialize!(cache, CssModulesCache);
-    let mut content_map = self.content_map.lock();
-
-    for (k, v) in cache.content_map {
-      content_map.insert(k, v);
+    if context.cache_manager.enable {
+      if let Some(cache) = context
+        .cache_manager
```
wre232114, line 201: This way is the same as `plugin_cache_load`, we want the cache key control by the module instead of the plugin, so we may need to add a API like `context.cache_manager.add_module_extra_info` the these info will be appended  when write the module cache.

When reading cache, we can use `handle_persistent_cached_module` hook to restore the cache back to the plugin

wre232114 (approved):

PR #2084, opened 2025-02-10
fix(core): incorrect schema valiation

nonzzz:
**Description:**

I'm working for visualization plugin, but i found a mistake and it prevent me

**BREAKING CHANGE:**


nonzzz:
@ErKeLost cc.

ErKeLost:
thanks

PR #2090, opened 2025-02-12
fix(schema): invalid server schema with types

fu050409:
**Description:**

**BREAKING CHANGE:**


packages/core/src/config/schema.ts
```diff
@@ -158,19 +158,13 @@ const serverSchema = z
...
-      ])
-      .optional(),
-    open: z.boolean().optional(),
     https: z.custom<SecureServerOptions>(),
-    cors: z.boolean().optional(),
-    appType: z.enum(['spa', 'mpa', 'custom']).optional(),
-    proxy: proxySchema,
-    strictPort: z.boolean().optional(),
+    origin: z.string().optional(),
+    protocol: z.enum(['http', 'https']).optional(),
```
ErKeLost, line 163: Schema mainly analyzes parameters passed by users to config, but we actually don't provide users with attributes like `hostname` or `protocol `. I think we may not need to verify these types. Is there any problem?

ErKeLost: I think we should modify the corresponding type, maybe the type caused a misunderstanding of the schema

fu050409: I see, I will try to omit these non-user configs in ts types.

ErKeLost:
I'll check this in a while.

ErKeLost (approved):

PR #2092, opened 2025-02-13
feat: add freeze_module hooks

ErKeLost:
**Description:**

**BREAKING CHANGE:**


crates/node/src/plugin_adapters/js_plugin_adapter/hooks/freeze_module.rs
```diff
@@ -0,0 +1,258 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(crate = "farmfe_core::serde", rename_all = "camelCase")]
+struct CompatiblePluginFreezeModuleHookParams {
+  module_id: ModuleId,
+  module_type: ModuleType,
+  content: String,
+}
+
+fn format_module_metadata_to_code(
```
wre232114, line 70: we can extract this method as a util which can be shared between different hook

wre232114 (approved):

PR #2096, opened 2025-02-16
fix: tree shake, hmr  and sourcemap issues

wre232114:
**Description:**

**BREAKING CHANGE:**


crates/compiler/src/update/regenerate_resources/mod.rs
```diff
@@ -85,26 +91,51 @@ pub fn render_and_generate_update_resource(
...
       let code = String::from_utf8(update_resources.resource.bytes).unwrap();
       let global_this = get_farm_global_this(
         &context.config.runtime.namespace,
         &context.config.output.target_env,
       );
+      let mut magic_string = MagicString::new(&code, None);
+
       // force re-register the affected modules when hmr
-      Ok(format!("{global_this}.{FARM_MODULE_SYSTEM}._rg=true;{code};{global_this}.{FARM_MODULE_SYSTEM}._rg=false;"))
+      magic_string.prepend(&format!("{global_this}.{FARM_MODULE_SYSTEM}._rg=true;"));
```
fu050409, line 108: Why not just using `&str` instead of `&String`?

PR #2098, opened 2025-02-19
[WIP]: Better Analysis tool 

nonzzz:
TL;DR

It's a long task and i still update.

### Check List

> Server 
- [x] module graph analysis. (Same as vite-bundle-analyzer)
- [ ] A simple Api server.
- [ ] evaluate plugin runtime.

> Client

- [] Refactor by react
- [] Redesign client UI

### Others

I gave up some of the integration logic. Instead, I used other implementations. Don't worry everything is go on.


.vscode/settings.json
```diff
@@ -2,6 +2,7 @@
   "editor.tabSize": 2,
   "editor.inlineSuggest.showToolbar": "always",
   "biome.enabled": true,
+  "editor.formatOnSave": true,
```
fu050409, line 5: Ah, I think we'd better not enable this feature in the workspace. It doesn't suit the development habits of all developers. Maybe you could move this setting to the VSCode global settings. WDYT?

nonzzz: I marked the task as wip in order to sync my config in different workspace. when the task is ready for check will be remove :D

Issue #2091, opened 2025-02-12
[Feature Request]: Refactor `create-farm` context


fu050409:
### What problem does this feature solve?

For now, we are using odd formating template to create new projects. Sometimes, some templates are broken and doesn't working fine because some project names injection (eg. tauri `Cargo.toml`).

We could refactor this crate and add a context, we can store all user inputs and functionally create a template project.

Also, we need to rewrite some farm config files inside the template projects, it's better to do this refactor. What's more, we can also add some extra selections for users to create different template due to this refactor.

### What does the proposed API look like?

see above

PR #2102, opened 2025-02-26
feat(create-farm): rewrite create farm

fu050409:
Resolved: #2091


ErKeLost:
I think we should try not to make things too complicated, such as `package manager` and `initializing git`. We need to allow users to quickly access the project. These tasks should be considered by higher-layer frameworks.

fu050409:
@ErKeLost Personally, I think the role of CLI tools is to save the user as much time as possible, and integrating tasks that may need to be performed in a CLI tool can help the user start a new project faster. In fact, in addition to making it clearer to maintain our CLI tools, the new architecture also makes it easier to expand the functionality of `create-farm` more quickly, and even create more subtasks for different package managers or template projects. For example, 他和 new version of `pnpm` has a security mechanism that causes Electron's dependency installations to fail, we can perform more actions or give more hints specifically for `pnpm`.

As you can see, special customizations are also done in `create-vue` for templates, such as JSX support, Vue Router configuration, Pinia and Vitest. So while this leads to making more choices when creating a project, it can be more convenient for the user. Besides, `create-tauri-app` also provides lots of choices for the user.

We could also follow this design, which would allow users to explore many of Farm's features earlier. In my vision, I would add some interesting optional generic plugins to `create-farm`, such as unpluin's imagemin, or our `@farmfe/plugin-compress`, and so on. The fact that we are currently facing the problem that many users don't find Farm's plugin ecosystem very quickly, which makes some users waste a considerable amount of time costing them a lot of money when using Farm, I think that these problems can be completely solved partially in `create-farm`, as this is something that almost every user starting Farm needs to do.

I followed the user needs oriented basis in the new refactoring to create less extraneous code while helping users with more tasks in the CLI tools. In the Git initialization task, I bind the `.gitignore` file to `.git`, since projects decoupled from Git don't seem to need a `.gitignore` file either. If you don't mind, maybe I can configure the Git initialization task and the tasks like configure biome or vitest in one task. We could consolidate the tasks in `create-vue` into one multi-select task, this can reduce the number of choices.

Of course, if you're sure we don't need to do that, I'll change the code later.

ErKeLost:
@fu050409 

On the one hand, since we are a basic framework, we should ensure user friendliness as much as possible, enabling students who are just starting to learn front-end development, for example, to use and understand the functions of this framework more quickly starter. 
On the other hand, based on many users' feedback or my own experience with some scaffolding tools and projects I've participated in before, users believe that the fewer configurations for this type of basic framework, the better. Maybe at first, people just want to test something and don't need to keep choosing configurations. This is my personal opinion.

fu050409:
@ErKeLost Thanks for the advice, I know what u are thinking abount but I still reserve my opinion.

I asked some working front-end engineers for their opinion in order to solve this problem, and it looks like the problem exists in both of the scenarios we mentioned. Some engineers prefer a strong scaffolding, but some of my friends (who may be using farm for the first time or are new to front-end development) object to the complexity this may bring.

Maybe we could use a compromise solution, such as letting users choose whether to go further or not, which would triage users: for those who want to test something or get up to speed with farm as soon as possible, they could choose no at this step, and for the other group of users who choose yes, our `create-farm` scaffolding could do the work for them more work for them.

ErKeLost:
I personally think this might not be what the create farm needs to do. If more powerful features are required, we can simply start a new scaffolding to handle these. This kind of issue is rather subjective. I also know that everyone will have different views. Everyone bases their opinions on their own experiences. We can ask for everyone's opinions.

fu050409:
I'll be moving forward with this one, and next I'll be placing all the other miscellaneous tasks after asking the user if they've decided to continue, as a way of accommodating the usage preferences of different user types.

PR #2140, opened 2025-04-02
fix: set `assetsFilename` options get not found resource

### LINKED ISSUE #2139: [Bug Report]: v2 set assetsFilename error
### Steps to reproduce

config

```ts
import { defineConfig, loadEnv } from "@farmfe/core";
import path from 'node:path'
import react from "@farmfe/plugin-react";

export default defineConfig({
  plugins: [
    react(),
  ],
  compilation: {
    input: {
      index: path.resolve(__dirname, "index.html"),
      base: path.resolve(__dirname, 'base.html'),
      about: path.resolve(__dirname, 'about.html'),
    },
    progress: false,
    persistentCache: false,
    sourcemap: false,
    output: {
      publicPath: "/aaa/",
      filename: '[ext]/[name]-[hash].[ext]',
      assetsFilename: 'assets/[name]-[hash].[ext]',
    },
    // partialBundling: {
    //   groups: [
    //     {
    //       name: "vendor-react",
    //       test: ["node_modules/"],
    //     },
    //   ],
    // },
  },
});

```

### Reproduce link

_No response_

Image

run browser get error

Image

### System Info

```Shell
mac
```

ErKeLost:
close #2139

### DIFF
--- a/packages/core/src/server/middlewares/resource.ts
+++ b/packages/core/src/server/middlewares/resource.ts
@@ -118,7 +118,9 @@ export function findResource(
     publicPath,
     req.url
   );
-  const normalizedPath = resourceWithoutPublicPath.replace(/^\//, '');
+  const normalizedPath = resourceWithoutPublicPath.startsWith('/')
+    ? resourceWithoutPublicPath.slice(1)
+    : resourceWithoutPublicPath;
   const resource = compiler.resource(normalizedPath);
 
   if (resource) {
--- a/packages/core/src/server/middlewares/resource.ts
+++ b/packages/core/src/server/middlewares/resource.ts
@@ -118,8 +118,8 @@ export function findResource(
     publicPath,
     req.url
   );
-
-  const resource = compiler.resource(resourceWithoutPublicPath);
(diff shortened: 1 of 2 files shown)


packages/core/src/server/middlewares/resource.ts
```diff
@@ -118,8 +118,8 @@ export function findResource(
     publicPath,
     req.url
   );
-
-  const resource = compiler.resource(resourceWithoutPublicPath);
+  const normalizedPath = resourceWithoutPublicPath.replace(/^\//, '');
```
fu050409, line 121:
```suggestion
  const normalizedPath = resourceWithoutPublicPath.startsWith('/') ? resourceWithoutPublicPath.slice(1) : resourceWithoutPublicPath;
```

Regular expression may cost more unexpected overheads, I think we should avoid using regex. WDYT?

ErKeLost: Performance may not be a top priority in the server, but you are right. Here I've thought of another method. Let me give it a try.

PR #2173, opened 2025-06-05
fix(core): Add origin check to the HMR server

### LINKED ISSUE #2168: Presence of Security Vulnerability
### Steps to reproduce

N/A

### Reproduce link

_No response_

Hello, I have found a security vulnerability in your code and would like to report it but it appears that you do not have a security policy set up or GiHub's vulnerability reporting system enabled. How would you like to receive the details about this? Thank you.

### System Info

```Shell
@farmfe/core: 1.6.7
```

R4356th:
**Description:**
Resolves #2168.

**BREAKING CHANGE:** Clients that previously used to connect to Farm's HMR server can no longer do so unless their requests are made from pre-configured origins.

## Summary by CodeRabbit

- **New Features**
  - Enhanced security for live updates by validating connection origins for WebSocket-based Hot Module Replacement.
- **Bug Fixes**
  - Improved handling of unauthorized WebSocket upgrade requests by closing invalid connections immediately.

### DIFF
--- a/.changeset/floppy-tires-worry.md
+++ b/.changeset/floppy-tires-worry.md
@@ -1,0 +1,5 @@
+---
+"@farmfe/core": patch
+---
+
+Added origin validation to HMR server
--- a/.changeset/floppy-tires-worry.md
+++ b/.changeset/floppy-tires-worry.md
@@ -3,3 +3,5 @@
 ---
 
 Added origin validation to HMR server
+
+BREAKING CHANGE: The HMR server now rejects all connections with unrecognized `Origin` headers. Clients need to update their configured ports and hosts if they want external apps to be able to connect to the HMR server.
--- a/packages/core/src/server/ws.ts
+++ b/packages/core/src/server/ws.ts
@@ -66,7 +66,8 @@ export default class WsServer implements IWebSocketServer {
     origins.push(`${protocol}://127.0.0.1:${port}`);
 
     // Add non-localhost origin
-    if (hostname && 
+    if (
+        hostname && 
         hostname.name && 
         hostname.name !== 'localhost' && 
         hostname.name !== '127.0.0.1'
--- a/packages/core/src/server/ws.ts
+++ b/packages/core/src/server/ws.ts
@@ -67,10 +67,10 @@ export default class WsServer implements IWebSocketServer {
 
     // Add non-localhost origin
     if (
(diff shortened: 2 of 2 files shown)


R4356th:
@wre232114, I have marked the change as minor because I suspect you don't want a major version bump just because of the unlikely possibility that someone maybe connecting to the HMR server for legitimate purposes.

fu050409:
Sometimes we may host development server on multiple network, @ErKeLost Take a look?

ErKeLost:
@R4356th  It seems that if you connect to the development server remotely through ssh or other means, you need to specify the host insert like farm.config.ts and server options host, right? I don't know what the current implementation of vite is like.

wre232114 (approved):

wre232114:
The lint check failed

R4356th:
> @R4356th It seems that if you connect to the development server remotely through ssh or other means, you need to specify the host insert like farm.config.ts and server options host, right? I don't know what the current implementation of vite is like.

Yes, the code should allow that. Why would you need to connect to the server through SSH though?

R4356th:
> The lint check failed

Apologies for that; it should pass now.

ErKeLost:
@R4356th  
The some team will let the development team use ssh to connect to the remote server and forward the remote service to the local in vscode. I have not tried this kind of development. I am not sure whether it is still accessing localhost.

fu050409:
@R4356th @ErKeLost If you means the remote development server of VSCode, the server connection will be redirected by VSCode to a localhost address. However, sometimes the server is limited which is only allowed to connect to server via the specific port, we cannot use VSCode remote anymore, we can only use a ssh connection to access server.

R4356th:
If I understand @fu050409's comment correctly, it should continue to work as long as the origin is properly configured.

ErKeLost:
@R4356th  If can point to localhost normally, I think there should be no problem. You may need to look at some CI error issues.

R4356th:
@ErKeLost, if you could please share an example then I could take a look at their implementation. Otherwise, you could check if those servers continue to work with this change before you merge it.
By the way, this is how webpack-dev-server implements the check, too, so if the servers you mention work with it (I do not know if that's the case) then I don't see any reason for them not to work with this change.

wre232114:
Thank you very much! It will be merged today

PR #2174, opened 2025-06-07
feat: add the packageManager for corepack

Nirvana-Jie:
**Description:**

Add the `packageManager` for `corepack` 

**BREAKING CHANGE:**
None

None


fu050409:
Some checks failed, may be u had forgot updating the lock file?

Issue #2191, opened 2025-06-23
[Feature Request]: No javascript option for react


xmh0511:
### What problem does this feature solve?

The default programming language is TypeScript. However, Farm does support JavaScript by replacing the default file extension with `jsx`. Please provide the option that allows us to use JavaScript as the major language for the project.

### What does the proposed API look like?

No suggestion

wre232114:
You can use use `.js` or `.jsx`, it's supported. What does `Farm does support JavaScript by replacing the default file extension with jsx` mean? Could you provide more details about the issue

xmh0511:
> You can use use `.js` or `.jsx`, it's supported. What does `Farm does support JavaScript by replacing the default file extension with jsx` mean? Could you provide more details about the issue

I meant that the CLI creating a `react` project defaults to creating `main.tsx`, which uses `TypeScript` as the default template. There is no option in the CLI to chose JavaScript as the default template for project. 

![Image](https://github.com/user-attachments/assets/36a4e6e3-e3ed-4af9-93fb-383358b2d132)

![Image](https://github.com/user-attachments/assets/3822e7c5-af11-465e-a4c7-c33ec27fb631)


Users must manually change, for example, `main.tsx` to `main.jsx` to use JavaScript in the project. Moreover, `typings.d.ts` in `src` and `farm.config.ts` hint that this project is using TypeScript.

faga295:
We can either rename the existing template to xxx-ts and then add js versions,
or we can let the user choose the language after selecting the template.IMO, the latter is better because it doesn't double the number of templates.

PR #2199, opened 2025-07-02
feat(create-farm): react js template

faga295:
**Description:**

**BREAKING CHANGE:**

#2191


ErKeLost:
@faga295  hi ,I think we can make different branches after we choose the framework. Templates in ts or js language may be better, and we can put them in v2-dev branch.

PR #2265, opened 2025-11-27
fix(plugin_css): handle empty 'url()' and improve performance

Nirvana-Jie:
## 📋 Description

This PR improves the CSS URL resolver in `plugin_css` by fixing empty URL handling and optimizing resource lookup performance.

## 🐛 Bug Fixes

- **Handle empty `url()` gracefully**: Previously, empty URLs could cause processing errors. Now returns `None` early to skip invalid URLs.
- **Remove panic risk**: Replaced `unwrap()` calls with safe `Option` handling. Missing resources now return `None` instead of panicking.

### Before
- **O(n) linear search** through all resources for each `url()` reference
- **Repeated computation** of `normalized_public_path` for every URL

### After
- **O(1) HashMap lookup** using pre-built reverse resource map (`ModuleId` → `resource_name`)
- **Pre-computed** `normalized_public_path` in constructor

## Summary by CodeRabbit

* **Refactor**
  * Unified CSS URL resolution and introduced normalized public-path handling for consistent asset rewriting.

* **Bug Fixes**
  * Ignore empty/whitespace url() values and reliably select the designated primary output when multiple resources exist.

* **New Features**
  * Added explicit primary marking for emitted resources to control which output is used for module assets.

* **Chores**
  * Updated example CSS/markup and switched emit API usage to constructor-style calls.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/crates/plugin_css/src/source_replacer.rs
+++ b/crates/plugin_css/src/source_replacer.rs
@@ -2,7 +2,7 @@ use farmfe_core::{
   config::AliasItem,
   module::{module_graph::ModuleGraph, ModuleId},
   plugin::ResolveKind,
-  resource::{Resource, ResourceOrigin},
+  resource::{Resource, ResourceOrigin, RESOURCE_META_PRIMARY_KEY},
   swc_common::DUMMY_SP,
   swc_css_ast::{AtRulePrelude, ImportHref, Rule, Str, Stylesheet, Url, UrlValue},
   HashMap,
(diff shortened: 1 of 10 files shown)


examples/css-url/src/main.css
```diff
@@ -1,6 +1,6 @@
 .main {
   color: red;
-  background: url(/@/logo.png);
+  background: url();
```
wre232114: Should add a new test case here, the original one is used to test alias

wre232114 (changes requested):

examples/css-url/src/main.css
Nirvana-Jie: got it

crates/plugin_css/src/source_replacer.rs
```diff
@@ -17,9 +17,9 @@ use crate::dep_analyzer::is_source_ignored;
 pub struct SourceReplacer<'a> {
   module_id: ModuleId,
   module_graph: &'a ModuleGraph,
-  resources_map: &'a HashMap<String, Resource>,
-  public_path: String,
   alias: Vec<AliasItem>,
+  normalized_public_path: String,
+  reverse_resource_map: HashMap<ModuleId, String>,
```
wre232114: A Module id may be mapped to multiple resources, we should use HashMap<ModuleId, Vec<String>> here. BTW, we should only map css related module resource map here.

